PR #7, opened 2015-04-07
Fixes cross_compilation variable substitution

gftg85:
The variable 'cross_compilation' was not substituted by 'configure', although configure is able to determine its value. This commit adds the variable 'cross_compilation' to 'ac_subst_vars'.

### DIFF
--- a/configure
+++ b/configure
@@ -621,6 +621,7 @@ with_dfp
 mzarch
 dfp_backend
 enable_decimal_float
+cross_compiling
 host_os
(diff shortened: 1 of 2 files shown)


tuliom:
@gftg85 , thanks for your contribution, but due to other problems in the project, we had to move away from using ldconfig in the Makefile.
Although your patch is right, it isn't necessary anymore because I had to remove that part of code in commit ID 32101191.

Issue #1, opened 2014-07-17
%f and %F cannot be used together in the same printf statement.


rzinsly:
Printf statements that uses %f and %Hf  didn't work, the values printed will be zero instead of the real values of the variables in %f and %Hf.

For example, this: printf("%f (long double) %Hf (_Decimal64)\n", number, dd)
Will generate this output: 0.000000 (long double) 0.000000 (_Decimal64)

From libdfp side, for a value with 32 bits the function __d32_va (./printf_dfp.c:300) is called with the wrong parameter va_list *ap when this combination occurs, so the function stores the wrong value (something like 0E-101), maybe this is caused by some bug in glibc side.

When %f and %Hf is used together ___printf_fp is called earlier in _IO_vfprintf_internal than other hook combinations that I observed, maybe this could indicate what is wrong.

tuliom:
As explained in pull request #8 , we can't reproduce this issue anymore.

PR #8, opened 2015-05-04
Issue #1 invalid?

gftg85:
Hi,

I wrote this patch set because I believe that issue #1 is not valid.

The first patch in the set adds several tests (actually copied from tests/test-printf.c) that pass two % modifiers to printf. The first of them is %Lf. All tests succeed.
NOTE: We probably do not want this patch added to the master branch, but I kept it in the patch set, because it shows that %Lf works with a lot of other % modifiers.

The second patch adds a wrong test case that reproduces the bug, and explains (in the log message) why I believe that issue #1 is invalid.

The third patch fixes the wrong test case.


rzinsly:
The issue #1  is not about using %f for long double. It is about using %f and %Hf in the same printf statement.
If you print a float with %f and a _Decimal32 with a %Hf in the same printf statement does it prints the correct values? If the answer is yes the issue #1 is indeed invalid.

gftg85:
I added another patch that prints with "%f %Hf", but passes to printf a float and a _Decimal32, respectively.

This test case also works, so can we conclude that issue #1 is invalid?

tuliom:
I rebased this pull request, changed its commit message and pushed it as commit 9dc973a2.

PR #11, opened 2015-07-10
Initial debianization

frediz:
Here is the packaging for debian I did.
Would you mind integrating this upstream ?


tuliom:
I noticed the debian scripts are not integrated into libdfp build system.
So, I merged this PR into branch deb so that we can work on this without affecting the master branch.

Thanks!!

PR #17, opened 2015-09-09
libdfp: powerpc: Fix optimized _Decimal128 implementations

zatrazz:
This patch changes the expected input register to _Decimal128 to
correct fr2 instead of fr0 (since ABI states _Decimal128 should
be passed in odd/even registers) on various powerpc implementations.

Signed-off-by: Adhemerval Zanella  <adhemerval.zanella@linaro.org>

### DIFF
--- a/sysdeps/powerpc/dfpu/fabsd128.c
+++ b/sysdeps/powerpc/dfpu/fabsd128.c
@@ -25,8 +25,10 @@ _Decimal128
 __fabsd128 (_Decimal128 x)
 {
   /* Half part os decimal128 constainst the sign bit at same position as
-     binary64, so the instruction works for both formats.  */
-  register _Decimal128 input asm("fr0") = x;
+     binary64, so the instruction works for both formats.  Also for 
+     _Decimal128 the ABI constraint it to odd/even register so it is
+     passed and returned in fr2/fr2.  */
+  register _Decimal128 input asm("fr2") = x;
   _Decimal128 ret;
   asm ("fabs  %1, %0\n"
        : "=f"(ret)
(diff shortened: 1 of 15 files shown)


sysdeps/powerpc/dfpu/fabsd128.c
```diff
@@ -25,8 +25,10 @@ _Decimal128
...
 {
   /* Half part os decimal128 constainst the sign bit at same position as
-     binary64, so the instruction works for both formats.  */
-  register _Decimal128 input asm("fr0") = x;
+     binary64, so the instruction works for both formats.  Also for 
+     _Decimal128 the ABI constraint it to odd/even register so it is
+     passed and returned in fr2/fr2.  */
+  register _Decimal128 input asm("fr2") = x;
   _Decimal128 ret;
   asm ("fabs  %1, %0\n"
```
pmur: Are the operands listed on line 33 backwards? I think that is the cause of the bug.

zatrazz: Indeed, the register bikeshedding are not really required with correct asm order. This:

```
_Decimal128
__fabsd128 (_Decimal128 x)
{ 
  /* Half part os decimal128 constainst the sign bit at same position as
     binary64, so the instruction works for both formats.  */
  _Decimal128 ret;
  asm ("fabs  %0, %1\n"
       : "=f"(ret)
       : "f"(x));
  return ret;
}
```

Generates the expect output:

```
$ objdump -d fabsd128.o

fabsd128.o:     file format elf64-powerpcle


Disassembly of section .text:

0000000000000000 <__GI___fabsd128>:
   0:   10 12 40 fc     fabs    f2,f2
   4:   20 00 80 4e     blr
        ...
```

zatrazz: I will update the patch.

PR #18, opened 2015-09-10
Master ppc fixes

zatrazz:


pmur:
LGTM. I verified the tests work on ppc64le/P8. Thank you for fixing that up.

PR #19, opened 2015-09-18
Fix configure when --with-cpu not supplied

pmur:
This will now attempt to use the newest CPU architecture available
from the compiler, and prevent a degenerate case whereby hardware
dfp is supported, but a specific PPC machine is not selected.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -247,6 +247,21 @@ AC_ARG_WITH([cpu],
     with_dfp=no ;;
   esac
 ])
+
+# If building for a soft dfp, we must explicitly
+# prevent the compiler from emitting hardware
+# instructions. Otherwise, unexpected behavior
+# will occur.
+#
+# Test to the toolchain to see if it supports the
+# option to disable hard dfp, and if so, disable it.
+# This should work for s390 and power.
+if test "$with_dfp" == "no"; then
+  cflags_tmp="$CFLAGS"
+  CFLAGS="$CFLAGS -mno-hard-dfp"
+  AC_COMPILE_IFELSE([AC_LANG_SOURCE([[]])],[],[CFLAGS="$cflags_tmp"])
+fi
+
 AC_SUBST(with_dfp)
 AC_MSG_RESULT($with_dfp)
 
--- a/configure.ac
+++ b/configure.ac
@@ -220,13 +220,29 @@ AC_ARG_WITH([cpu],
 ],[
   case "$host_cpu" in
   powerpc*)
+    # Test successively newer processors and select the
+    # best implementation.
     AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
-      #ifdef _ARCH_PWR6
-      # define SUCCESS 1
-      #else
+      #ifndef _ARCH_PWR6
       # error "Support for hardware dfp isn't available"
       #endif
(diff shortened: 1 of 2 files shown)


pmur:
This fixes issue #19

tuliom:
This patch is interesting, but I believe we should have some code first to force soft dfp when with-cpu isn't specified.
Only after that we could have a patch that tries to automatically get the cpu value like this one.

configure.ac
```diff
@@ -220,13 +220,37 @@ AC_ARG_WITH([cpu],
...
     AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
       #ifdef _ARCH_PWR6
       # define SUCCESS 1
       #else
       # error "Support for hardware dfp isn't available"
       #endif
-      ]])], [with_dfp=yes], [with_dfp=no]) ;;
+      ]])], [with_dfp=yes; submachine=power6], [with_dfp=no])
+    AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+      #ifdef _ARCH_PWR6X
```
tuliom: Shouldn't you test for _ARCH_PWR6 ?

configure.ac
```diff
@@ -220,13 +220,37 @@ AC_ARG_WITH([cpu],
...
+      # define SUCCESS 1
+      #else
+      # error "Support for hardware dfp isn't available"
+      #endif
+      ]])], [with_dfp=yes; submachine=power6x], [])
+    AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+      #ifdef _ARCH_PWR7
```
tuliom: Do we really need to test for POWER7 and POWER8?
They both set _ARCH_PWR6 and _ARCH_PWR7.

pmur: Yes, the order here matters. We test for increasingly newer processors until we fail.

Suggestions for a more canonical way to test for these values? Ideally, we might just run one test which would print out the correct values based on the supplied macros.

tuliom:
> Yes, the order here matters. We test for increasingly newer processors until we fail.

I wasn't really referring to the order, but if it was really required to test for POWER7 and POWER8. But now I see I was still blindly following the old behavior that didn't use to change `submachine`.
Let me put it in a different way: to find the value of `with_dfp`, you only need to test for POWER6.

> Suggestions for a more canonical way to test for these values? Ideally, we might just run one test which would print out the correct values based on the supplied macros.

You could encapsulate the duplicated code, so that you could have something like this:
```
CMACRO_IFELSE([_ARCH_PWR6],
              [with_dfp=yes; submachine=power6],
              [with_dfp=no]);
CMACRO_IFELSE([_ARCH_PWR6X],
              [submachine=power6x], );
CMACRO_IFELSE([_ARCH_PWR7],
              [submachine=power7], );
...
```

pmur:
I've updated the patch to simplify the macro tests as suggested, and added a new patch to pass -mno-hard-dfp when building for soft_dfp on power/z targets.

tuliom:
Both patches LGTM in overall.  You may merge it yourself after:
 - Adding a signed-off-by tag to all of them (according to https://github.com/libdfp/libdfp/blob/master/CONTRIBUTING#L50-L52)
 - Improving the output of ```configure --help``` to explain the usage of ```--with-cpu=no```.

pmur:
Merged with suggested changes.

PR #20, opened 2015-09-21
Fix edge case with strtod*

### LINKED ISSUE #3: test failure
On ppc64el, there a failure on test-strtod : 
$ cat test-strtod.out
104 Error:   Expected: "+0,000,000,000,000,000E-398"
             Result:   "+5,000,000,000,000,000E-398"
    in: ./tests/test-strtod.c:119.

Found 1 failures.

pmur:
An edge case overlooked values a multiple of 10 higher than what could be represented by the _Decimal* types.

This check is enhanced to set errno in these cases. A few extra test cases are added to catch the same bug with 32 and 128b implementations. And the lm2lmd_to_c table is increased to better handle this if it slips through.

This should fix #3


tuliom:
LGTM.
It's just missing a Signed-off-by.

pmur:
Merged with suggested changes.

PR #23, opened 2015-10-20
Remove a workaround from make-release.sh

tuliom:
Since configure now attempts to automatically detect the cpu,
make-release.sh doesn't need to force a specific CPU while configuring
libdfp.

Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.vnet.ibm.com>


pmur:
LGTM

PR #24, opened 2015-12-02
__int128 <-> _Decimal128 DPD Conversion Support

pmur:
BID _Decimal128 -> __int128 support is still TODO.


pmur:
I reworked the conversion routines to support dd and df support, and removed a bit of duplicate code.

BID is failing tests unexpectedly, some comparisons seem to be getting evaluated wrong. (i.e -1 > 0 returning true). I need to look into that.

sysdeps/dpd/Makefile
```diff
@@ -1 +1 @@
-libdfp_files += dpd-private binpowof10 decpowof2
+libdfp_files += dpd-private binpowof10 decpowof2 ticonstants fixtdti fixunstdti
```
tuliom: I can't build these files on ppc32.

tuliom:
This PR is breaking the check-localplt.out on ppc64.

pmur:
So, TImode probably won't ever work on a 32 bit platform. The build system needed a bit of monkeying as this causes us to support another ABI without TImode.

Unfortunately, this makes the tedious task of maintaining the ABI check 2x as time consuming. 

I'm not going to add a configuration option to control this, as it will become increasingly complicated to maintain as float128 will also depend on this.

pmur:
Fixed a typo symbol-hack.h causing unsdddi to get linked to unsddti, causing a test failure with soft-dfp.

pmur:
Thanks.

I agree, the ctz/clz bug *should* have been picked up by the test suite. However, the test suite does not do much (any?) testing of proper exception handling.

pmur:
Fixed the comment and merged via a rebase.

PR #28, opened 2015-12-18
Another round of TI support

pmur:
Fix missed comments from previous review.
Fix warnings when building in the source tree.
Fix for GCC < 4.9


tuliom:
LGTM.

PR #35, opened 2015-12-22
Final set of patches for DPD TI mode support

pmur:
This implements the float*ti{dd,sd} functions, and fixes a double rounding bug uncovered in trunctdsd2 while testing whether the default rounding mode was working as expected.

### DIFF
--- a/base-math/floattitd.c
+++ b/base-math/floattitd.c
@@ -24,6 +24,8 @@
 
 #include <dfpacc.h>
 
+#include "convert_helpers.h"
+
 /*
  * Convert a signed 128-bit binary integer into nearest representable
  * IEEE754R 128-bit Densely Packed Decimal Floating-point (DFP).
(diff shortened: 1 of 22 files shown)


base-math/floattitd.c
```diff
@@ -95,7 +120,7 @@ __BACKEND_ (floattitd) (INT128 a)
     {
       /* If abs(a) is less than 63-bits we can convert the signed
        * value directly.  */
-      result = (long long) a;
+    result = (INPUT_TYPE64) a;
```
gftg85: What happens when the return type (RET_TYPE?) is a Decimal32?
This is possible, right?

gftg85: Oh, just saw the comment on convert_helpers.h. Nevermind my question. :)

tuliom:
LGTM

gftg85:
@pmur, thanks for your explanations. Your patch looks good to me.

pmur:
Merged. Thanks!

PR #40, opened 2016-01-15
Add hack to fix tests building with older GCC

pmur:
This seems to mitigate the issue without introducing added complexity.


sysdeps/soft-dfp/dpd/numdigits.h
```diff
@@ -214,7 +214,10 @@ FUNC_D (left_justify) (DEC_TYPE x)
   __get_digits_d32(x, digits, NULL, NULL, NULL, NULL);
 #elif _DECIMAL_SIZE == 64
 # define decfield dd
-  union ieee754r_Decimal64 mask = { .di = {
+  /* The 'static const' is a hack to prevent an ancient version
+     of GCC from transforming this NaN and removing its payload,
```
tuliom: Could you mention the GCC version, please?

LGTM with that change.

pmur:
Pushed with clarified comments. Thanks!

PR #41, opened 2016-01-27
Lots of optimizations

pmur:
The focus of these are primarily to improve conversion speed by avoiding redundant operations, redundant checks, and supplying more optimal integer extraction routines.

This is at least a 2x speedup for  td -> df conversions, and likely more for other routines.

This also adds efficient BID TI conversions too.

Tested for ppc64, ppc64le, ppc64le-sdfp, ppc64-sfp, ppc32, ppc32-sdfp, x86-64 x86, and RHEL6 ppc BE targets.


pmur:
@tuliom I've updated the bottom commit to include a generic fenv_libdfp.h, and it is always included in convert.h, so the test for the fegetenv_register macro behaves as expected.

Issue #43, opened 2016-02-17
FTBFS on s390x configured --with-cpu=z10


xnox:
When configuring libdfp on s390x with z10 CPU setting, the package fails to build from source:

```
$ make
find: ‘./sysdeps/s390/dfpu/libdfp-test-ulps’: No such file or directory
find: ‘./sysdeps/s390/fpu/libdfp-test-ulps’: No such file or directory
find: ‘./sysdeps/s390/libdfp-test-ulps’: No such file or directory
find: ‘./sysdeps/dpd/libdfp-test-ulps’: No such file or directory
VPATH=./sysdeps/s390/dfpu:./sysdeps/s390/fpu:./sysdeps/s390:./sysdeps/dpd:./sysdeps/generic:./decNumberMath/:./ieee754r/:./base-math:./:./tests
Processing Makefiles: ./sysdeps/dpd/ ./decNumberMath/ ./ieee754r/ ./base-math
gcc  -DSHARED -O2 -fpic -g -march=z10  -mzarch -MT floattitd.o -MD -MP -MF floattitd.d -c ./base-math/floattitd.c -D__STDC_DEC_FP__=200704L -D__STDC_WANT_DEC_FP__=1 -DOPTION_EGLIBC_LOCALE_CODE=1 -std=gnu99 -D_ISOC99_SOURCE -W -Wall -Wwrite-strings -Wstrict-prototypes -Wmissing-prototypes -Wold-style-definition -Wmissing-format-attribute -Wno-long-long -include ./config.h -include ./include/libdfp-symbols.h  -I./include  -I./dfp  -I./dfp/decimal  -I./libdecnumber/dpd  -I./libdecnumber  -I./sysdeps/s390/dfpu  -I./sysdeps/s390/fpu  -I./sysdeps/s390  -I./sysdeps/dpd  -I./sysdeps/generic  -I./decNumberMath/  -I./ieee754/  -I./base-math/  -I.  -I. -o floattitd.o
./base-math/floattitd.c:27:29: fatal error: convert_helpers.h: No such file or directory
compilation terminated.
Makefile:241: recipe for target 'floattitd.o' failed
make: *** [floattitd.o] Error 1
```

pmur:
I suspect this should be fixed by MR41.

I'm working on getting my zVM running again to verify on this platform.

xnox:
I can try to validate that MR.

pmur:
[...]

PR #44, opened 2016-02-18
Get s390 working again

pmur:
make, and make check should work against the soft-dfp and z10 targets.

There is quite a bit of room for optimization on z. Based on what I saw, all the functions in convert_helpers.h should be fairly trivial to implement.


pmur:
This should fix #43, and many other things I fixed along the way.

pmur:
@tuliom No good deed goes unpunished. This is the "minimal" amount of work to get the s390 targets working again.

tuliom:
Did you test if make-release.sh works after these changes?

pmur:
It worked on PPC64LE. Though, for libdfp it isn't much more than running configure/make/make check.

tuliom:
LGTM.

pmur:
After reviewing, comparing, contrasting, and testing, I've merged patches 5 6 8 and 9. The others are duplicated by Stefan, and he took a few extra steps for the soft-fp and 31 bit variants too. Thus why I've merged most of his, as they are very similar if not the same as most of these.

PR #45, opened 2016-03-02
Libdfp on s390.

stliibm:
Hi,
this patch series enables building libdfp on s390.
The testsuite passes for soft- and hardware decimal floating point on s390x (64bit) and s390 (31bit).
Bye Stefan


pmur:
Hi Stefan,

Thank you for helping get S390 back into shape.

Coincidentally, I just went through a similar process a couple of weeks ago. Merge request #44 seemed to get everything passing for me. I'll need to figure what, if any, needs to be merged
between the two.

I have a few general questions since I went through a similar process not so long ago:

How does s390 handle BFP <-> DFP conversions? When I did the work, the compiler was inserting
special code to handle them. The BFP conversions in libdfp usually only accurate to about 1ulp.

Is TFmode for s390 exclusively IEEE 128b floating point?

Some of the existing inline ASM I found also seemed to override to an incorrect rounding mode. Most of the digit counting tricks seem to rely on round to zero (truncate), and the __int128 conversions use "round to prepare for shorter precision".

pmur:
Hi Stefan,

One more question, I'm pretty sure you already have it, but can you confirm your FSF copyright assignment?

Thanks,
Paul

stliibm:
Yes I have FSF copyright assignment for project glibc. (See the entry Stefan Liebler at https://sourceware.org/glibc/wiki/MAINTAINERS#Maintainers_.28developers.29_for_libc) The libdfp project is not explicitely listed in FSF. Is this okay?

stliibm:
First:
Sorry, I missed this pull-request :-)

Second:
S390 uses only 128bit long double.
For bfp <-> dfp conversion, gcc emits the pfpo "perform floating-point operation" instruction.

Remaining:
I compared the pull-request 44  to 45:

* "Ensure hidden prototypes are used" == "S390: Use direct calls of __isnand64|128 instead of extra PLT references."

* "s390: Cleanup numdigits" == "S390: Get rid of compile warnings with hw-dfp."

* "s390: add convert_helper.h" vs "S390: Add convert_helpers.h implementation for hw-dfp.":
My patch implements combine_and_truncd32|64 in the same way,
by setting "Round to prepare for shorter" around the combination and the critical add-instruction.
In addition, converting from _Decimal128 to _Decimal64 is done in this rounding mode, too.
Otherwise I got:
floatunsti32 (340282349999999999999999999999999999999) in: /home/stli/libdfp/src/tests/test-float.c:242
1   Error:   Expected: "3.40282300000000000000000000000000e+38"
              Result:   "3.40282400000000000000000000000000e+38"
in: /home/stli/libdfp/src/tests/test-float.c:146

For GCCs < 5, the direct conversion from _Decimal128 to _Decimal32 in libdfp
can be changed to use this rounding mode. GCCs >= 5 emits two load rounded instructions
without this rounding-mode for _Decimal128 -> _Decimal64 conversion.
Thus combine_and_truncd32 should do the _Decimal128 -> _Decimal64 conversion in this rounding mode.
(See comment to "S390: Fix casting from _Decimal32|128 to _Decimal128|32 with hw-dfp.")




* "s390: Add soft-fp for d32 <-> d128 conversions" vs "S390: Fix casting from _Decimal32|128 to _Decimal128|32 with hw-dfp.":
My patch generates code to use the ldetr, lxdtr instructions for d32 -> d128 conversion
and ldxtr, ledtr instructions for d128 -> d32 conversion.
Starting with gcc 5, __dpd_extendsdtd2 and __dpd_trunctdsd2 aren't called,
but these instructions are emitted directly.




* " Fix tests when long double is IEEE128" vs ""
yours: #else  /* __LDBL_MANT_DIG__ == 106 or 113 */
mine: #elif __LDBL_MANT_DIG__ == 106 || __LDBL_MANT_DIG__ == 113
I think yours is better.

Why do you use #if __LDBL_MANT_DIG__ == 113 ... #else ... around xxtf_tests?
Isn't it possible to specify the long double literals with L suffix - e.g 12.345L - on power?
Or can we at least do something like that:
-once at top of file:
#if __LDBL_MANT_DIG__ == 113
#define TF(x) x##L
#else
#define TF(x) x
#endif
-for every tf-literal:
TF(12.345)




* "s390: Fix left_justify" vs 
"S390: Use correct instruction in left_justifyd128() function with hw-dfp."
and "S390: Fix left_justifyd32() with hw-dfp."
The effect is similar: for _Decimal32, there are only _Decimal64 instructions,
thus we need the _Decimal64 instructions for left_justifyd32,
but then the adjusting of digits for d32 has to use "ahi %1,-15\n\t"
instead of "ahi %1,-6\n\t", too?



* "s390: Add missing test files" vs ...
- sysdeps/s390/dfpu/libdfp-test-ulps:
  Why do you list tests with 0 ulps?
  For log10, I only need "decimal64 1" entry for "test log10 3.141592...",
  but not the "decimal128 1" entry for "test log10 2.718..".
  See "S390: New libdfp-test-ulps file with hw-dfp."

- sysdeps/s390/fpu/libdfp-test-ulps:
  I thought sysdeps/soft-dfp/libdfp-test-ulps is used, thus I have not created
  a specific one for s390 with sw-dfp. The soft-dfp file contains the same entries
  except the 0 ulps entries.
  I saw "src/tests/gen-libdfp-ulps.py src/sysdeps/soft-dfp/libdfp-test-ulps -o libdfp-test-ulps.h"
  while running "make check".

- sysdeps/s390/dfpu/libdfp.abilist:
  This file differs to the power-dfpu-version due to missing hidden_def (...)
  in base-math files (See "libdfp: Add hidden definition for base_math symbols.").
  Power uses his own files in sysdeps/power/... which have the hidden_def.
  That's the same on Intel, too.
  See "S390: Add new libdfp.abilist for hw-dfp."
  In addition to this abilist file, I've created one for 31bit where no int128
  data-types are available. See "S390: Add new libdfp-no-ti.abilist for 31bit sw-/hw-dfp."

- sysdeps/s390/fpu/libdfp.abilist:
  This patch file is identical to mine.
  See "S390: Add new libdfp.abilist for sw-dfp.".
  In addition to this abilist file, I've created one for 31bit where no int128
  data-types are available. See "S390: Add new libdfp-no-ti.abilist for 31bit sw-/hw-dfp."

- sysdeps/s390/dfpu/localplt.data:
  As you already mentioned in your comment: It's due to optimize-sibling-calls.
  See "S390: Fix "Missing required PLT reference: libdfp.so: register_printf_dfp"."






* "s390: Properly override long double functions" vs nothing:
Which gcc version did you use?
My tested gccs use the PFPO "perform floating-point operation" to convert between
_Decimalxx and long double with hard-dfp. Thus I do not need these conversion methods at all.
For soft-dfp, those extend|trunc functions are called and the libdfp ones are power-specific.
Thus I did the patch "S390: Fix conversion from|to decimal combined with long double with sw-dfp.".
But now I think there is no need for them, too. Libgcc has it's own extend|trunc functions.
If the extend|trunc*.c files are empty, the testsuite passes, except for the check-abi-test.
Thus we those functions in libdfp.
But the following leads to segmentation fault with sw-dfp, because the function is called recursively.
_Decimal128 __dpd_extendtftd(long double d)
{
  return d;
}

pmur:
Thanks for taking a look at MR#44 too.

Regarding test-bfp-conversions:

They make a number of implicit assumptions about the behavior of ibm128, which don't carry over to true IEEE 128. The tdtf test for PPC just ignores the lower double, and compares the high part. Thus why 12.345 vs 12.345L in that case.

pmur:
For the tests, including dfpacc.h should declare all the implicit functions provided by gcc. Anyhow, those fixes can wait for another patch. No need to fix them now.

pmur:
The ULPs file was copied from PPC, my intuition being they should be identical.

Though, in hindsight, there are a few optimized functions for PPC which are not specific to PPC. I.e logd{64,32} and expd{32,64}. I'll need to figure out how to move them around so s390 can pick them up too.

pmur:
My test system is running RHEL 7.1 with binutils from Fedora 22. I believe Andreas set it up for me last year.

pmur:
Pending verification, I plan to commit all patches except 2 3 7 and 16 here, and from my MR, merge commits 5 6 8 and 9. My fixes for left_justify, fixing test-bfp-conversions, and overriding the long double conversion functions.

In the next few months, IEEE128 support will be coming into libdfp. So I'm a little hesitant about embedding too much into processor specific subdirs.

Again, thank you for your contribution.

pmur:
I've gone ahead and pushed all the commits except those stated above. I tested for z196, z196 with soft-dfp, and the equivalent -m31 variants. All tests passed on all the known platforms to support.

stliibm:
Thanks for committing these patches.

I have still two open question:
1) Why have you commited the
"return (_Decimal32) (_Decimal64)in;"
in sysdeps/s390/dfpu/trunctdsd2.c?

Didn't you mention "The conversion from D128 -> D64 should be done in "round to prepare for shorter precision". I fixed a similar bug in PPC." ?

I've added the inline assembly, which use "Round to prepare for shorter" for the D128 -> D64 conversion.
(See [comment](https://github.com/stliibm/libdfp/commit/6daad608034bd20467464065084f3ea6898edf09#commitcomment-16457175))
Gcc will be fixed, too.

The testcase test-float passes because this D128 -> D64 conversion is done explicitely in this rounding mode (See convert_helpers.h: combine_and_truncd32)


2) The fast_truncdxx functions in convert_helpers.h do net recognize inexact exception. But you've mentioned, that they should recognize them.
This is done by changing M4 flag of fidtr/fixtr instruction from 4 to 0:

```
/* Truncate quickly.  No need to validate input.  */
static inline _Decimal128
fast_truncd128 (_Decimal128 a)
{
  /* Load FP Integer with round toward 0 with inexact-exception.  */
  asm ("fixtr %0,5,%0,0\n\t" : "+f" (a));
  return a;
}

static inline _Decimal64
fast_truncd64 (_Decimal64 a)
{
  /* Load FP Integer with round toward 0 with inexact-exception.  */
  asm ("fidtr %0,5,%0,0\n\t" : "+f" (a));
  return a;
}
```

Issue #31, opened 2015-12-19
Conversions incorrect on overflow


jsm28:
At least some libdfp conversions between decimal and binary types wrongly always produce an infinity for obviously overflowing results, without taking account of the rounding mode (the relevant rounding mode is that for the radix of the target type, not the source type).  For example, the attached test converting 1e1000DL to double produces (on POWER8 little-endian):

inf

but the expected result, as with the libgcc conversions, is:

0x1.fffffffffffffp+1023
[overflow.txt](https://github.com/libdfp/libdfp/files/67519/overflow.txt)

racardoso:
Closed via  bc38bdf

PR #59, opened 2017-08-23
Fix incorrect conversion on overflow.

racardoso:
This commit fixes incorrect conversions between decimal and binary types that
always produce an infinity for obviously overflowing results, without taking
account of the rounding mode (task #31).

### DIFF
--- a/base-math/truncdddf.c
+++ b/base-math/truncdddf.c
@@ -28,6 +28,9 @@
 #define DEST 64
 #define NAME trunc
 
+#include <float.h>
+#include <fenv.h>
+
 #include "dfpacc.h"
 #include "convert.h"
 #include "convert_helpers.h"
@@ -49,7 +52,19 @@ CONVERT_WRAPPER(
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);
-	    return SIGNBIT(a) ? -INFINITY : INFINITY;
+
+	    switch (fegetround())
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -DBL_MAX : DBL_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : DBL_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -DBL_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
+	      }
 	  }
 	else if (exp < -BINPOWOF10_LIMIT)	/* Obvious underflow.  */
 	  {
--- a/base-math/truncddsf.c
+++ b/base-math/truncddsf.c
@@ -28,6 +28,9 @@
 #define DEST 32
 #define NAME trunc
 
+#include <float.h>
+#include <fenv.h>
+
 #include "dfpacc.h"
 #include "convert.h"
 #include "convert_helpers.h"
@@ -47,7 +50,19 @@ CONVERT_WRAPPER(
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);
-	    return SIGNBIT(a) ? -INFINITY : INFINITY;
+
+	    switch (fegetround())
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -FLT_MAX : FLT_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : FLT_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -FLT_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
+	      }
 	  }
 	else if (exp < -39)	/* Obvious underflow.  */
 	  {
--- a/base-math/trunctddf.c
+++ b/base-math/trunctddf.c
@@ -28,6 +28,9 @@
 #define DEST 64
 #define NAME trunc
 
+#include <float.h>
+#include <fenv.h>
+
 #include "dfpacc.h"
 #include "convert.h"
 
@@ -48,7 +51,19 @@ CONVERT_WRAPPER(
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);
-	    return SIGNBIT(a) ? -INFINITY : INFINITY;
+
+	    switch (fegetround())
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -DBL_MAX : DBL_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : DBL_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -DBL_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
+	      }
 	  }
 	else if (exp < -BINPOWOF10_LIMIT)		/* Obvious underflow.  */
 	  {
--- a/base-math/trunctdsf.c
+++ b/base-math/trunctdsf.c
@@ -28,6 +28,9 @@
 #define DEST 32
 #define NAME trunc
 
+#include <float.h>
+#include <fenv.h>
+
 #include "dfpacc.h"
 #include "convert.h"
 #include "convert_helpers.h"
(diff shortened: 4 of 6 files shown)


racardoso:
I am working on the underflow problem too. But I am willing to hear any reviews on this before. This is my first PR on this project.

base-math/truncdddf.c
```diff
@@ -49,7 +52,20 @@ CONVERT_WRAPPER(
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);
-	    return SIGNBIT(a) ? -INFINITY : INFINITY;
+
+		/* Default result when trap is disable - IEEE754 */
```
gftg85: I did not understand this comment...  Did you mean something in the lines of:

Correctly round the result according to the current rounding direction mode.

base-math/truncdddf.c
```diff
@@ -49,7 +52,20 @@ CONVERT_WRAPPER(
...
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -DBL_MAX : DBL_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : DBL_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -DBL_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
```
gftg85: In lines 59 to 67, replace 8 spaces with 1 tab.

base-math/truncddsf.c
```diff
@@ -38,16 +41,29 @@ CONVERT_WRAPPER(
 	_Decimal64 a_norm;
 	long long mant;
 	int exp, sexp;
-	
+
```
gftg85: I'd prefer that this change, which is not related to the rounding direction mode problem, not be integrated.

base-math/truncddsf.c
```diff
@@ -38,16 +41,29 @@ CONVERT_WRAPPER(
...
 	a_norm = getmantandexpd64 (a, &exp, 16, 1e16DD);
-	
+
```
gftg85: Likewise.

base-math/truncddsf.c
```diff
@@ -38,16 +41,29 @@ CONVERT_WRAPPER(
...
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -FLT_MAX : FLT_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : FLT_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -FLT_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
```
gftg85: In lines 57 to 65, replace 8 spaces with 1 tab.

base-math/trunctddf.c
```diff
@@ -48,7 +51,20 @@ CONVERT_WRAPPER(
...
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -DBL_MAX : DBL_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY : DBL_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -DBL_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
```
gftg85: In lines 58 to 66, replace 8 spaces with 1 tab.

base-math/trunctdsf.c
```diff
@@ -38,15 +41,28 @@ CONVERT_WRAPPER(
 	SRC_TYPE a_norm;
 	long long mant;
 	int	exp, sexp;
-	
+
```
gftg85: Same comment about unrelated changes.

base-math/trunctdsf.c
```diff
@@ -38,15 +41,28 @@ CONVERT_WRAPPER(
...
+	      {
+	        case FE_TOWARDZERO:
+	          return SIGNBIT(a) ? -FLT_MAX : FLT_MAX;
+	        case FE_DOWNWARD:
+	          return SIGNBIT(a) ? -INFINITY :FLT_MAX;
+	        case FE_UPWARD:
+	          return SIGNBIT(a) ? -FLT_MAX : INFINITY;
+	        case FE_TONEAREST:
+	        default:
+	          return SIGNBIT(a) ? -INFINITY : INFINITY;
```
gftg85: In lines 56 to 64, replace 8 spaces with 1 tab.

base-math/truncdddf.c
racardoso: @gftg85  That weird comment I forgot to remove. That's was for remind me that POWER ISA 2.07 says (pag. 125) we should adjust the exponent of the intermediate result and adjust the round result if the overflow exception is enable. If it's disable we should do as I implement. It's not clear to me if in this case "obvious overflow" when cast to double if we need to adjust the exponent and return as "Normal Number" or just do what I am doing.

base-math/truncdddf.c
racardoso: @gftg85 I was following how it's implemented on the rest of the file. On my vim it appears (set list) as one tab and spaces. But ok I'll check that.

racardoso:
Tasks: 

- [ ] Add a test to avoid regressions.

base-math/truncdddf.c
gftg85: Hrm, you are right.  Please disregard this comment.

It seems that there isn't a coding style being followed and the relevant section in README.developer is in TODO state.

Sorry about the inconvenience.

base-math/truncdddf.c
gftg85: Section 4.4.3 of POWER ISA 2.07 only refers to operations with binary floating-point formats.  Section 5.5.10.4 refers to operations with decimal floating-point formats.  None of them describe how conversions between these floating-point types work. Moreover, they are specific to POWER.

On the other hand, ISO-IEC TR 24732, in section 6.2, describes how a C implementation of decimal floating-point type should make the conversions between floating-point types.  Based on that, I think that your implementation is correct.

base-math/truncddsf.c
racardoso: I agree. But notice that for base-math/trunctddf.c there's no tab on those lines. Maybe I should do another PR to fix that.

base-math/truncddsf.c
racardoso: Likewise.

base-math/trunctdsf.c
racardoso: Same thing about differences between trunc*** files. Is this tab incorrect?

gftg85:
Your white-space fixes seem reasonable.  My only concern is that they have nothing to do with the overflow problem, so I'd like to see these changes removed from this patch.

If you want to fix the white-space changes in another commit, that's fine by me.

base-math/truncddsf.c
```diff
@@ -40,14 +43,26 @@ CONVERT_WRAPPER(
 	int exp, sexp;
 	
 	a_norm = getmantandexpd64 (a, &exp, 16, 1e16DD);
-	
+
```
gftg85: Thanks for removing the other occurrences of this change.
This one slipped in...  Can you remove it, too?

tests/test-cast-to-overflow.c
```diff
@@ -0,0 +1,91 @@
+
```
gftg85: This empty line is not necessary and I don't see it in other files.
Could you remove it?

tests/test-cast-to-overflow.c
```diff
@@ -0,0 +1,91 @@
...
+   General Public License version 2.1.
+   The Decimal Floating Point C Library is distributed in the hope that
+   it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+   the GNU Lesser General Public License version 2.1 for more details.
+   You should have received a copy of the GNU Lesser General Public
+   License version 2.1 along with the Decimal Floating Point C Library;
+   if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+   Suite 330, Boston, MA 02111-1307 USA.
+   Please see libdfp/COPYING.txt for more information.  */
```
gftg85: This license block has a weird format.
Could you re-write it so that it looks similar to the other files?

tests/test-cast-to-overflow.c
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+  for (i=0; i<d128to32_size; ++i)
+    {
+      float ret;
+      fesetround (d128to32[i].round_mode);
+      ret = (float) d128to32[i].set;
+      _VC_P (__FILE__, d128to32[i].line, d128to32[i].expect, ret, "%a");
+    }
+
```
gftg85: Do we want to test for long double, too?

base-math/truncddsf.c
racardoso: Oops. I will remove it.

tests/test-cast-to-overflow.c
racardoso: Yes.

tests/test-cast-to-overflow.c
racardoso: I did a ctrl+c ctrl+v on the license from another file? What's wrong? Indentation?

tests/test-cast-to-overflow.c
racardoso: Isn't long double and double the same on POWER?

tests/test-cast-to-overflow.c
gftg85: At a glance it seems different.

I see that there are empty lines splitting the paragraphs in the other files.  Maybe that's the only difference.

PS: I just noticed that the year is 2016.  Please change it to 2017.

tests/test-cast-to-overflow.c
gftg85: No.  It might have been in the past, but it is not anymore.  On POWER, long double uses two 64-bits registers to get more precision (without changing the range, i.e.: no changes in the exponent size).

https://en.wikipedia.org/wiki/IBM_Floating_Point_Architecture#Extended-precision_128-bit

racardoso: Turns out we found a new bug. Conversion to long double returns a NaN instead Infinity. I've opened a new task (#60) to fix this problem. It's a different kind of problem since on that case it's not related to the rounding mode but to the conversion itself. (It's something before the rounding mode switch).

gftg85: OK.  Thanks for the new bug report.
You don't need to address that in this pull request.

gftg85:
Thanks for the new patch with the fixes...

Before I can approve this, could you rebase it and turn these three commits into a single commit or into two commits (one for the fix and the other for the test case)?

Issue #32, opened 2015-12-19
Conversions incorrect on underflow


jsm28:
At least some libdfp conversions between decimal and binary types wrongly always produce a zero for obvious underflowing results, without taking account of the rounding mode (the relevant rounding mode is that for the radix of the target type, not the source type). For example, the attached test converting 1e-1000DL to double produces (on POWER8 little-endian):

0x0p+0

but the expected result, as with the libgcc conversions, is:

0x0.0000000000001p-1022
[underflow.txt](https://github.com/libdfp/libdfp/files/67523/underflow.txt)

racardoso:
Closed via  c5f353e

jsm28:
This fix looks from for FE_TONEAREST mode, when conversion of a value with absolute value less than or equal to half the least subnormal in the target type should produce +/- 0, not +/- the least subnormal.

racardoso:
@jsm28 You're right. I will fix it here.

PR #61, opened 2017-09-12
Fix incorrect conversion on underflow.

racardoso:
This commit fixes incorrect conversions between decimal and binary types that
always produce an +/-0.0 for obviously overflowing results, without taking
account of the rounding mode (task #32).

### DIFF
--- a/base-math/truncdddf.c
+++ b/base-math/truncdddf.c
@@ -28,6 +28,9 @@
 #define DEST 64
 #define NAME trunc
 
+#include <float.h>
+#include <fenv.h>
+
 #include "dfpacc.h"
 #include "convert.h"
 #include "convert_helpers.h"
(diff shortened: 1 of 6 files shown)


base-math/truncdddf.c
```diff
@@ -55,7 +58,19 @@ CONVERT_WRAPPER(
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_UNDERFLOW|FE_INEXACT);
-	    return SIGNBIT(a) ? -0.0 : 0.0;
+
+	    switch (fegetround())
+	      {
+	        case FE_TONEAREST:
+	          return SIGNBIT(a) ? -__DBL_DENORM_MIN__ : __DBL_DENORM_MIN__;
```
gftg85: IEEE 754, section 4.3.2, states that:

> roundTowardPositive, the result shall be the format’s floating-point number (possibly +∞) closest
to and no less than the infinitely precise result

It is my understanding that the "**format's floating point number**" indeed includes the subnormal set.  So I believe your implementation is correct.

Issue #34, opened 2015-12-19
Conversions wrongly clear already-raised "inexact" exception


jsm28:
At least some libdfp conversions between decimal and binary types wrongly clear the "inexact" exception when it was raised before the conversion; the exception flag bits are sticky and should not be cleared except by a few functions such as feclearexcept.  For example, the attached test prints

inexact wrongly cleared

instead of the expected

inexact still raised as expected

that it does with the libgcc conversions.
[testinex2.txt](https://github.com/libdfp/libdfp/files/67526/testinex2.txt)

racardoso:
The problem is in this line https://github.com/libdfp/libdfp/blob/c5f353ed6cef0571b8056a8bee3d4db742722faf/base-math/trunctddf.c#L105 it always clear `FE_INEXACT` no matter if we force a exception or if we produce a inexact result. Comment this line make the test pass and I didn't see any side effects. I don't know why this line was added. I am working on it.

Issue #33, opened 2015-12-19
Conversions missing "inexact" exception


jsm28:
At least some libdfp conversions between decimal and binary types wrongly fail to raise the "inexact" exception for inexact results.  For example, the attached test prints

inexact not raised

instead of the expected

inexact raised as expected

(libgcc conversions via strings have the same issue because glibc strtod fails to raise "inexact", which is a strtod bug; this bug report is only about the libdfp issue).
[testinex.txt](https://github.com/libdfp/libdfp/files/67525/testinex.txt)

racardoso:
Related with task #34.

racardoso:
I am checking if I have to clean this inexact if the result is a normal number. If I have to, I will have to check any exceptions before the code and then check if the number is not a subnormal in the end.

PR #67, opened 2018-03-20
Fix raised inexact exception wrong cleanup.

racardoso:
This commit removes a code that force clear a raised inexact exception. Issues
(#33 #34).

### DIFF
--- a/tests/test-inexact-exception.c
+++ b/tests/test-inexact-exception.c
@@ -1,0 +1,103 @@
+/* Test Inexact Exception.
+
+   Copyright (C) 2018 Free Software Foundation, Inc.
+
+   This file is part of the Decimal Floating Point C Library.
+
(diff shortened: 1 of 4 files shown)


racardoso:
@gftg85 
1. About extendddtf I am not sure but truncdfsd.c I forgot (in fact I need to check truncsdsf.c and truncdfsd.c because they seems incomplete - no underflow/overflow check in either files)
2. Yes and I will adapt those tests. Thanks for the suggestion.

tests/test-inexact-exception.c
```diff
@@ -0,0 +1,103 @@
+/* Test trucate cast Decimal -> Binary overflow.
+
+   Copyright (C) 2017 Free Software Foundation, Inc.
```
tuliom: s/2017/2018/

PR #68, opened 2018-03-27
Allow use of -Werror

gftg85:
Some warnings currently prevent the use of `-Werror` in CFLAGS in libdfp.  This patch set cleans up the warnings, thus allowing the following command to complete successfully:

```
CFLAGS="-Wall -Werror -O2" ./configure
make && make check
```

People testing libdfp in continuous integration would be able to use it and get cleaner logs.

### DIFF
--- a/configure
+++ b/configure
@@ -4385,12 +4385,11 @@ int
 main ()
 {
 
- static int foo;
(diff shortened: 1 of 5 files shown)


tuliom:
Merged.

PR #70, opened 2018-04-16
Adapt the code for glibc 2.28

tuliom:
Replace the following references:
 - _IO_size_t with size_t;
 - _IO_FILE with FILE.

Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>


racardoso:
@tuliom are you replace or removing the reference? Why they are no longer needed?

Issue #71, opened 2018-05-15
s390x: Failing tests test-cast-to-overflow / test-cast-to-underflow


stliibm:
On s390x (with hardware dfp) the test test-cast-to-overflow fails with:
3   Error:   Expected: "0x1.fffffffffffffp+1023"
              Result:   "inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:49

4   Error:   Expected: "-0x1.fffffffffffffp+1023"
              Result:   "-inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:50

5   Error:   Expected: "0x1.fffffffffffffp+1023"
              Result:   "inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:51

8   Error:   Expected: "-0x1.fffffffffffffp+1023"
              Result:   "-inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:54

11  Error:   Expected: "0x1.fffffep+127"
              Result:   "inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:61

12  Error:   Expected: "-0x1.fffffep+127"
              Result:   "-inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:62

13  Error:   Expected: "0x1.fffffep+127"
              Result:   "inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:63

16  Error:   Expected: "-0x1.fffffep+127"
              Result:   "-inf"
in: /home/stli/libdfpDir/src/tests/test-cast-to-overflow.c:66

Found 8 failures.

and test-cast-to-underflow with:
5   Error:   Expected: "0x0.0000000000001p-1022"
              Result:   "0x0p+0"
in: /home/stli/libdfpDir/src/tests/test-cast-to-underflow.c:51

6   Error:   Expected: "-0x0.0000000000001p-1022"
              Result:   "-0x0p+0"
in: /home/stli/libdfpDir/src/tests/test-cast-to-underflow.c:52

10  Error:   Expected: "-0x0.0000000000001p-1022"
              Result:   "-0x0p+0"
in: /home/stli/libdfpDir/src/tests/test-cast-to-underflow.c:56

11  Error:   Expected: "0x0.0000000000001p-1022"
              Result:   "0x0p+0"
[...]

Issue #60, opened 2017-09-04
Conversion incorrect for long double when overflow.


racardoso:
`long double` type should, depending on the rounding mode, return `INFINITY` when overflow instead, it returns NaN. For example, the attached test converting `1e1000DL` to `long double` produces (on POWER8 little-endian):

```
#include <fenv.h>                                                               
#include <float.h>                                                               
#include <math.h>

int                                                                             
main (void)                                                                     
{ 
  volatile long double;
  volatile _Decimal128 a = 1e1000DL;    
  fesetround (FE_TONEAREST);                                                    
  b = (long double)  a;
  return b != INFINITY; 
} 
```
The expected result should be `inf` instead `nan`.

racardoso:
Fixed in 9c45074fcba2bf9bc7517ea80ae6b37c82da322e

PR #72, opened 2018-05-15
Fix some issues on s390x (see referenced issue for each commit)

stliibm:


racardoso:
In fact there's I suggest you move the issue number from the title to the commit message. In some cases you're already mention the issue on the message. Also this will make the commit https://github.com/libdfp/libdfp/pull/72/commits/df707e3875b3ec20f4f713126906fab8557cd6dc title fit on one line.

racardoso:
Besides the commit message everything looks good to me.

stliibm:
I've changed the commit messages and pushed them to my branch. The file diffs are the same.
The issues are now referenced in the messages instead of the title and the commit message
of "Add more under/overflow tests for dfp to bfp conversion." is shortened.

racardoso:
@stliibm LGTM

stliibm:
Is this pull-request okay? Is there needed anything else?

racardoso:
@tuliom @gftg85

tests/test-cast-to-overflow.c
```diff
@@ -39,55 +39,234 @@ typedef struct
...
+ {__LINE__, FE_TONEAREST, 1e6000DL, INFINITY},
+ {__LINE__, FE_TONEAREST, -1e6000DL, -INFINITY},
+ {__LINE__, FE_TONEAREST, 1e4933DL, INFINITY},
+ {__LINE__, FE_TONEAREST, -1e4933DL, -INFINITY},
+#if __LDBL_MANT_DIG__ == 106
+ {__LINE__, FE_TONEAREST, 1.797693134862316e308DL, INFINITY},
+ {__LINE__, FE_TONEAREST, -1.797693134862316e308DL, -INFINITY},
+#else
+ {__LINE__, FE_TONEAREST, 1.189731495357232e4932DL, INFINITY},
+ {__LINE__, FE_TONEAREST, -1.189731495357232e4932DL, -INFINITY},
```
gftg85: I understand that these numbers are LDBL_MAX and -LDBL_MAX for IBM long double and IEEE long double (binary128), respectively, but I wonder if there's a way to express them in a way that makes this fact more clear...

Maybe with macros:
```
#if __LDBL_MANT_DIG__ == 106
# define LDBL_MAX_IN_DECIMAL 1.797693134862316e308DL
#else
# define LDBL_MAX_IN_DECIMAL 1.189731495357232e4932DL
#endif
(etc.)
```

Or maybe with an explicit cast (would this achieve the desired goal of this test?):
```
 {__LINE__, FE_TONEAREST, (_Decimal128) LDBL_MAX, INFINITY},
 {__LINE__, FE_TONEAREST, (_Decimal128) -LDBL_MAX, -INFINITY},
```
(then there would be no need to test for `__LDBL_MANT_DIG__`)

tests/test-cast-to-overflow.c
```diff
@@ -39,55 +39,234 @@ typedef struct
...
+#endif
+ {__LINE__, FE_TOWARDZERO, 1e6000DL, LDBL_MAX},
+ {__LINE__, FE_TOWARDZERO, -1e6000DL, -LDBL_MAX},
+ {__LINE__, FE_TOWARDZERO, 1e4933DL, LDBL_MAX},
+ {__LINE__, FE_TOWARDZERO, -1e4933DL, -LDBL_MAX},
+#if __LDBL_MANT_DIG__ == 106
+ {__LINE__, FE_TOWARDZERO, 1.797693134862316e308DL, LDBL_MAX},
+ {__LINE__, FE_TOWARDZERO, -1.797693134862316e308DL, -LDBL_MAX},
```
gftg85: This will fail on POWER, with the following message:

```
7   Error:   Expected: "0x1.fffffffffffff7ffffffffffff8p+1023"
              Result:   "0x1.fffffffffffffp+1023"
in: ./tests/test-cast-to-overflow.c:74
```
(similar problems will occur on different lines)

But that's probably due to the fact that _Decimal128 to long double conversions haven't been patched for subnormals (similarly to what was done in commit ID 4f67ffa3c65fed967ad23b8d2aea9921f0bd1cc0, but now for function `trunctdtf`).  Anyhow, that's not your problem and it shouldn't block this pull request, in my opinion.

tests/test-cast-to-overflow.c
gftg85: This comment, as well as the other one, shouldn't necessarily block this pull request, it's intended as a question (i.e.: are these constants indeed LDBL_MAX, FLT_MAX, etc. expressed as decimal constants? or is there some detail about them?).

stliibm: The numbers are not LDBL_MAX. Instead they are a little bit greater than LDBL_MAX.
I've defined some macros for those values to get rid of "#if __LDBL_MANT_DIG__"
and to point out, that those values are not LDBL_MAX.
See the further commit "Make test-cast-to-overflow.c more readable." on my branch. The original commit is unchanged.

tests/test-cast-to-overflow.c
stliibm: Yes there are some test-fails on power due to the current implementation (also in commit ID 4f67ffa).
Currently trunctdtf.c includes extendddtf.c. I think this is not correct?
Instead there has to be an implementation like trunctddf.c as you've already mentioned.
In issue #71 racardoso said, that he will work on those issues.

But I have anoter question:
Is the __LDBL_MAX__ value correct?
And if yes, why?
See the attached test:
ldbl_tst =     0x1.ffffffffffffffffffffffffff800000p+1023
is greater than
__LDBL_MAX__ = 0x1.fffffffffffff7ffffffffffff800000p+1023

[tst-ldbl-max.c.tar.gz](https://github.com/libdfp/libdfp/files/2206385/tst-ldbl-max.c.tar.gz)

racardoso: @gftg85 @stliibm long double is not working at all https://github.com/libdfp/libdfp/issues/60 also there's other reported issues related to long double such as unimplemented functions.

tests/test-cast-to-underflow.c
```diff
@@ -39,63 +39,461 @@ typedef struct
...
+ {__LINE__, FE_TONEAREST, -1e-325DL, -0.0},
+ {__LINE__, FE_TONEAREST, 2.4703282292062326e-324DL, 0.0},
+ {__LINE__, FE_TONEAREST, -2.4703282292062326e-324DL, -0.0},
+ {__LINE__, FE_TONEAREST, 2.4703282292062327e-324DL, 0.0},
+ {__LINE__, FE_TONEAREST, -2.4703282292062327e-324DL, -0.0},
+ {__LINE__, FE_TONEAREST, 2.470328229206232720882843964341105e-324DL, 0.0},
+ {__LINE__, FE_TONEAREST, -2.470328229206232720882843964341105e-324DL, -0.0},
+ {__LINE__, FE_TONEAREST, 2.470328229206232720882843964341107e-324DL,
+  __DBL_DENORM_MIN__},
+ {__LINE__, FE_TONEAREST, -2.470328229206232720882843964341107e-324DL,
```
racardoso: @stliibm I expect this test to return 0.0. "When conversion of a value with absolute value less than or equal to half the least subnormal in the target type should produce +/- 0, not +/- the least subnormal." 

Half of the least subnormal (DBL_DENORM_MIN) = 2.4703282292062327e-324.

Since the function in trunctddf.c (see https://github.com/libdfp/libdfp/blob/master/base-math/trunctddf.c#L47) truncates the number by 17 digits the result (a_norm) of 2.470328229206232720882843964341107e-324DL is
24703282292062327 exp=323 so the result here is -0.0 since 24703282292062327 is equal half of min subnormal.

I am not sure if the correct behavior will be checking if the result is less or equal half of subnormal before truncate (maybe creating a DL with the half subnormal and compare).

tests/test-cast-to-overflow.c
gftg85: Yes, there are bugs with the long double conversion functions __in libdfp__.  While I understand that this needs to be fixed, it shouldn't block this pull request.  That's what I meant with my comment.

As for Stefan's question about `ldbl_tst` ability to be greater than LDBL_MAX, there's an explanation to it in this bugzilla issue ([PR 70117](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70117)).  The explanation to the problem is interspersed with another discussion related to the `isinf` function, but comment [#5](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70117#c5) explains that GCC uses a 106-bit mantissa, and comment [#7](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70117#c7) suggests that this is probably not going to change.  `ldbl_tst` has a 107-bit mantissa and will compare greater than LDBL_MAX.

tests/test-cast-to-underflow.c
stliibm: The gcc macro is defined as:
define __DBL_DENORM_MIN__ ((double)4.94065645841246544176568792868221372e-324L)
This value is not truncated to 17 digits.

If running test-cast-to-underflow on s390x which uses hardware instruction pfpo for converting between the floating-point-types,
all tests are passing (if either gcc is fixed or the rounding-mode is set with fe_dec_setround. See issue #71).

I've used the attached test to see how conversion from double to float is handling this underflow case:
half_of_flt_denorm_min=7.0064923216240854e-46
test_float_precision=9.8765430450439453
test_double_precision=9.8765432310987648634181823581457138
================================================
convert_bfp64_to_bfp32():
in = 7.0064923216240846e-46 = 0x1.fffffffffffffp-151
out= 0.0000000000000000e+00 = 0x0p+0

convert_bfp64_to_bfp32():
in = 7.0064923216240854e-46 = 0x1p-150
out= 0.0000000000000000e+00 = 0x0p+0

convert_bfp64_to_bfp32():
in = 7.0064923216240869e-46 = 0x1.0000000000001p-150
out= 1.4012984643248171e-45 = 0x1p-149

According to ieee 754-2008:
4.3 Rounding-direction attributes
Rounding takes a number regarded as infinitely precise and, if necessary, modifies it to fit in the
destination’s format while signaling the inexact exception, underflow, or overflow when appropriate (see 7).
Except where stated otherwise, every operation shall be performed as if it first produced an intermediate
result correct to infinite precision and with unbounded range, and then rounded that result according to one
of the attributes in this clause.

One attribute can be round to nearest with ties to even.
[test-underflow.txt](https://github.com/libdfp/libdfp/files/2209817/test-underflow.txt)

racardoso: > The gcc macro is defined as:
define DBL_DENORM_MIN ((double)4.94065645841246544176568792868221372e-324L)
I did not say the macro was truncated. What is being truncated is the result of:
'''
/* Get mantissa with 17 significant digits, and normalized exponent.  */
a_norm = getmantandexpd128 (a, &exp, 17, 1e17DL);
'''
In fact I was using the wrong macro before. I got the macro using 'gcc -dM -E - < /dev/null' but I've used it on my laptop (a x86_64).

racardoso: I forgot to mention: I am not blocking this PR I still think that we should approve it.

PR #75, opened 2018-08-15
Add support for POWER9 and others

tuliom:
1. Improve current tests to validate the return code from *printf() upon successful calls.
2. Add support for POWER9.
3. Remove the ugly messages from find that appear when running make.


tuliom:
I'm merging this now.

PR #76, opened 2018-08-22
[RFC] Partially fix some of ov(und)erflow issues.

racardoso:
This commit is just intend to start a discussion about the underflow/overflow
issues on libdfp. As pointed out my last commit was returning some incorrect
results. The problem is that I don't think we should do a switch case and
test if the conversion results on a underflow/overflow. The division should
at end of code should round convert and return the correct result. Althought
we may return early if it's obvious overflow/underflor - if the exponent is
bigger than the max exponent (overflow) or less than min subnormal exponent
(underflow).

Those functions basically truncates the number and returns the exponent and a
SRC_TYPE (Decimal) after that it converts this to a DI integer to get the
mantissa as a long long int.

The function who truncates the number also sums 1 to the exponent. So,
for instance if the number is converted to a double an the exponent is
308 (MAX_DBL_10_EXP) the function will return 309.

My questions is if we should test if exp is bigger than 308 or if it
is bigger than 309 (MAX_DBL_10_EXP + 1). We test it after or before
the number was truncated?

This commit partially fix some errors on the new tests but, there's still
failures.


stliibm:
I've see the following fails (numbers are line numbers) in test-cast-to-underflow.c:
d128to64[]:
-FE_TONEAREST: (143), (144), 151, 153,
-FE_TOWARDZERO: 169, 171,
-FE_DOWNWARD: 186, 188,
-FE_UPWARD: 203, 205,

d128to32[]:
-FE_TONEAREST: (219), (220), 227, 229, 231, 232, (233), (234),
-FE_TOWARDZERO: 247, 249, 251, 252,
-FE_DOWNWARD: 264, 266, 268, 269,
-FE_UPWARD: 281, 283, 285, 286,

 d64to64[]:
-FE_TONEAREST: (295), (296),

d64to32[]:
-FE_TONEAREST: 349, 350,
The numbers in parantheses marks fails which are passing if your patch is applied and are failing without your patch.

The maximum value of a 64bit binary floating point is (1-2^-53)x2^1024 ~ 1.8x10^308.
Thus an exponent > 308 is an obvious overflow and should be handled by the special case.

Yes you are right, the "not so obvious" over-/underflow cases could be handled by the multiplication/division and the truncation from long double to double (in case of e.g. truncdddf.c).
Those instructions should return the correct value dependent on the current rounding mode and should raise the correct exceptions.
But this will only work if all digits are involved.

E.g. if the _Decimal128 value 2.470328229206232720882899999999999e-324DL is casted to double,
the value should not be truncated to 17 digits: 2.4703282292062327e-324DL.
The first value should be casted to __DBL_DENORM_MIN__, the latter one to 0.0.

As a "long long mant" can't store all 34 digits of a _Decimal128 value, the mantissa should be split into two long long values.
But we only need the second part of the digits in the "not so obvious" underflow cases?!

PR #79, opened 2019-02-21
Fix overflow test failures for flt32 conversion.

racardoso:
This commit fixes the two failures on cast-to-overflow test which
converts a Decimal32 to a float (binary 32). The range should be 38
or FLT_MAX_10_EXP not 39.

Signed-off-by: Rogerio Alves <rcardoso@linux.ibm.com>

### DIFF
--- a/base-math/trunctdsf.c
+++ b/base-math/trunctdsf.c
@@ -45,7 +45,7 @@ CONVERT_WRAPPER(
 	a_norm = getmantandexpd128 (a, &exp, 15, 1e15DL);
 	/* Handle obvious overflow and underflow to avoid going beyond the
 	   bounds of the exponent table.  */
-	if (exp > 39)		/* Obvious overflow.  */
+	if (exp > FLT_MAX_10_EXP)		/* Obvious overflow.  */
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);


base-math/trunctdsf.c
```diff
@@ -45,7 +45,7 @@ CONVERT_WRAPPER(
 	a_norm = getmantandexpd128 (a, &exp, 15, 1e15DL);
 	/* Handle obvious overflow and underflow to avoid going beyond the
 	   bounds of the exponent table.  */
-	if (exp > 39)		/* Obvious overflow.  */
+	if (exp > __FLT32_MAX_10_EXP__)		/* Obvious overflow.  */
```
gftg85: This macro is only available since GCC 7.  Is that a problem for libdfp?  Should we use 38, instead?

gftg85:
For the record, these are the errors that are fixed with this merge request:
```
43  Error:   Expected: "inf"
              Result:   "0x1.fffffep+127"
in: ./tests/test-cast-to-overflow.c:138

44  Error:   Expected: "-inf"
              Result:   "-0x1.fffffep+127"
in: ./tests/test-cast-to-overflow.c:139
```

base-math/trunctdsf.c
racardoso: Maybe I should create this macro instead

racardoso:
> For the record, these are the errors that are fixed with this merge request:
> 
> ```
> 43  Error:   Expected: "inf"
>               Result:   "0x1.fffffep+127"
> in: ./tests/test-cast-to-overflow.c:138
> 
> 44  Error:   Expected: "-inf"
>               Result:   "-0x1.fffffep+127"
> in: ./tests/test-cast-to-overflow.c:139
> ```

Yes

base-math/trunctdsf.c
gftg85: If you do, you have to protect the definition with a check for the compiler version, otherwise, your definition will conflict with the definition provided by GCC.  Also, notice that you may not use FLT32_MAX_10_EXP (without the leading and trailing underscores), because that name is reserved by ISO TS-18661.

tuliom: You may also use `__FLT_MAX_10_EXP__` or even `FLT_MAX_10_EXP`, which is provided by `float.h` as required by ISO C11.

racardoso: Ok I'll maybe use `__FLT_MAX_10_EXP__` instead create a macro then.  But, once again is this macro available in GCC < 7?

racardoso:
> Which architectures are affected by this modification?
x86, s390 and POWER are the only architectures supported by libdfp. The exponent is defined bu IEEE and should not vary from one architecture to another

> Why? Because IEEE single precision floating point cannot represent a number decimal exponent equals to 39.

`__FLT_MAX_10_EXP__ = 38`
`__FLT_MAX__ 3.40282346638528859811704183484516925e+38F`

Any exponent bigger than 38 will give an overflow

racardoso:
@gftg85 @tuliom bump. (forced push)

base-math/trunctdsf.c
```diff
@@ -45,7 +45,7 @@ CONVERT_WRAPPER(
 	a_norm = getmantandexpd128 (a, &exp, 15, 1e15DL);
 	/* Handle obvious overflow and underflow to avoid going beyond the
 	   bounds of the exponent table.  */
-	if (exp > 39)		/* Obvious overflow.  */
+	if (exp > __FLT_MAX_10_EXP__)		/* Obvious overflow.  */
```
gftg85: I think you should use `FLT_MAX_10_EXP`, since this macro is defined in ISO C, as pointed out by Tulio, and you already include `float.h` anyway.

racardoso: @gftg85 Ok I understand that I could use `__FLT_MAX_10_EXP__` or `FLT_MAX_10_EXP`. But, ok I replaced that on my last push. Also changed flt32 to float on the commit message

PR #81, opened 2019-03-20
Fix expoenent range check on Decimal64 to Float conversion

racardoso:
The range should be 38 or FLT_MAX_10_EXP not 39 to check if the exponent
is out of range on Decimal64 to float (binary 32) conversion.

Signed-off-by: Rogerio Alves <rcardoso@linux.ibm.com>

### DIFF
--- a/base-math/truncddsf.c
+++ b/base-math/truncddsf.c
@@ -46,7 +46,7 @@ CONVERT_WRAPPER(
 
 	/* Check for values that would overflow the exponent table, which
 	   would be obvious overflow and underflow.  */
-	if (exp > 39)		/* Obvious overflow.  */
+	if (exp > FLT_MAX_10_EXP)		/* Obvious overflow.  */
 	  {
 	    if (DFP_EXCEPTIONS_ENABLED)
 	      DFP_HANDLE_EXCEPTIONS (FE_OVERFLOW|FE_INEXACT);


gftg85:
Do you understand why the current code doesn't cause a failure in `test-cast-to-overflow.c`'s lines 178 and 179:

```
 {__LINE__, FE_TONEAREST, DEC64_GREATER_FLT_MAX, INFINITY},
 {__LINE__, FE_TONEAREST, -DEC64_GREATER_FLT_MAX, -INFINITY},
```

as it used to cause in the Decimal128 to float conversion in lines 138 and 139:

```
 {__LINE__, FE_TONEAREST, DEC128_GREATER_FLT_MAX, INFINITY},
 {__LINE__, FE_TONEAREST, -DEC128_GREATER_FLT_MAX, -INFINITY},
```

If so, could you describe this in the commit message?

Also, in the commit message, please fix the typos: `expoenent` -> `exponent`, and `binary 32` -> `binary32`.

racardoso:
> Do you understand why the current code doesn't cause a failure in `test-cast-to-overflow.c`'s lines 178 and 179

This test pass because the current code is testing for `exp > 39` and `DEC64_GREATER_FLT_MAX` is equal to `3.402823567797337e38DD` since the exponent is 38 after the normalization `exp` is  equal 39 and the code doesn't fall on switch. Bellow in `truncddsf.c` line 89 it multiplies result by `BINPOWOF10` table to adjust and by coincidence the result is `infinity` which is the correct result.

On Decimal128 the test fails before because this is not the correct behavior. Instead fall on the switch on a obvious overflow, since `exp == 39` the overflow was not detected so, it's was performing a multiply and adjust the result on the lines bellow. But, in that case it was returning the minimum denormal instead infinity which is wrong. All obvious overflow/underflow cases must be treated before.

> If so, could you describe this in the commit message?

Yes . But I must resume the message above. I guess it may be a quite confuse

> Also, in the commit message, please fix the typos: `expoenent` -> `exponent`, and `binary 32` -> `binary32`.

Ok.

gftg85:
>Yes . But I must resume the message above. I guess it may be a quite confuse

I still don't see, in the commit message, something mentioning that this patch does not fix any failures in any of the test cases.  I understand that the patch does the right thing, but I suggest that such a comment be added to the commit message, otherwise we loose this information (i.e. the fact that this patch does not solve a test case failure).

How about the following message (the explanation is a little bit different than yours, so please check that it is correct):

Conversions from Decimal128 to float (handled by base-math/trunctdsf.c) correctly check that the extracted exponent is greater than 38 (FLT_MAX_10_EXP) and, if not, call the overflow handling code.  However, before commit ID d5695775f259, they compared the exponent against 39, which caused a couple of test cases to fail (test-cast-to-overflow.c lines 138 and 139).

Similarly, conversions from Decimal64 to float (handled by base-math/truncddsf.c) incorrectly compare the extracted exponent against 39 before selecting between the overflow handling code or the regular code.  Although incorrect, this check does not cause any visible errors (nor any test case failures, as was the case for Decimal128 conversions), but that is only by chance.

For the Decimal128 case, the code in base-math/trunctdsf.c asks for 15 bits of mantissa when calling getmantandexpd128.  Afterwards, it subtracts 15 from the exponent before accessing the powers of 10 table to perform its calculation, then ends up in the test case failures (before the fix in commit ID d5695775f259).  On the other hand, base-math/truncddsf.c asks for 16 bits of mantissa and subtracts 16 from the exponent, thus accessing a different entry from the table and *accidentally* producing the correct result.

>All obvious overflow/underflow cases must be treated before.

Yes, and this is the correct rationale for this patch.

racardoso:
@gftg85 I've just made a new PR fixing the commit message. I just edited some parts of your suggestion

PR #86, opened 2019-09-04
Install COPYING.txt in inst_docdir.

stliibm:
Also install COPYING.txt to /usr/share/doc/dfp/COPYING.txt.
Then this file can be specified as %license in a spec file for an rpm package.

Signed-off-by: Stefan Liebler <stli@linux.ibm.com>


tuliom:
This pull request has just been merged.
Thanks!

Issue #25, opened 2015-12-03
tf functions broken with BID encoding


pmur:
The existing long double <-> _Decimal128 conversion functions assume IBM long double. This is not always correct.

pmur:
This also applies when TFmode maps to IEEE128 too.

pmur:
This should be fixed. Specifically, x86 DFP is not providing x86 TF variants.

PR #88, opened 2019-10-07
Fixup x86 conversion routines

pmur:
Some time ago, s390 worked around similar issues by doing radix
conversions via string conversion. Simple, slow, and usually
more correct.

Shuffle the Implies around such that these can be used for other
architectures. Similarly, move the symbol information to avoid
exposing TFmode or XFmode variants where not implemented. Similarly,
split the Version files to avoid arch specific long double details
from polluting common Version files.

Likewise, s390 and ppc both use TFmode to represent long double
operands. x86 uses XFmode, and TFmode for _Float128. Today libdfp
makes no attempt to support directly support _Float128, thus x86
users get broken long double conversions. Therefore, we must update
the macros to use XFmode in place of TFmode.

Likewise, fixup FP_LLOGB* macros. They are wrong for x86.

Finally, cleanup error message printed for long int checks so the
output formats match. Notably, test-cast-underflow tests still fail.
There are accuracy issues at the extreme ends of conversion which
have been uncovered the by the tests added to verify s390 hardware
conversions.

This also resolves Issue #25 by replacing TF functions with correct XF variants.


pmur:
One more fixup. the x86 sysdep dir should be i386 for consistency with libc.

pmur:
This was needed in my pile of build system patches. It is now merged. Thank!

PR #90, opened 2019-10-16
Build system cleanup

pmur:
This is the first set of patches to start refactoring the libdfp build system to something which kind of looks like glibc, with some minor fixes along the way.

This has been tested against ppc64le (hard and soft), i386, x86_64 and build tested for s390 and s390x.


pmur:
@stliibm if you have a chance, would you be able to test the s390{,x} variants? Am I correct to assume -mzarch is only needed for hard-dfp?

tests/test-cast-to-underflow.c
```diff
@@ -34,6 +34,14 @@
...
 #include "scaffold.c"
 
+/* TODO: libdfp conversion routines are not correct to 1ulp.
+         It would seem s390 hardware routines are more accurate. */
+#if !defined(__s390x)
+#define XFAIL_CONVERSION 1
+#else
+#define XFAIL_CONVERSION 0
+#endif
+
```
stliibm: XFAIL_CONVERSION should be "0" if we run on s390/s390x and if hardware dfp is available but not if we use software dfp.
In either case, the macro "__s390x" is never defined, but instead the following macros are defined on:
-s390 (31bit): __s390__
-s390x (64bit): __s390__ and __s390x__
Thus you can detect s390 (31 or 64bit) if __s390__ is defined.
The information if hardware dfp is used is currently not available via a macro definition from gcc or e.g. in config.h.
But we know it in configure.ac as with_dfp is set to yes or no.

stliibm:
@pmur I've successfully build and tested the following variants (including a hack for test-cast-to-underflow; see extra comment):
-s390x (64bit) with hw dfp
-s390x (64bit) with sw dfp
-s390 (31bit) with hw dfp
-s390 (31bit) with sw dfp

Yes, the dfp instructions are only available with zarch. This is the default if building for s390x (64bit) but not for s390 (31bit).

pmur:
@stliibm would you be able to rerun these patches on z? The tests should be correctly marked only to run on z w/dfp.

stliibm:
Hi @pmur . I've tested your branch. The "#if !XFAIL_CONVERSION" tests in test-cast-to-underflow.c are now executed on s390x with hw-dfp and are skipped on s390x without hw-dfp.
Thanks.

I've also build and run the testuite with different gcc's from 4 to 9 on s390x. Despite the known fails with older gccs due to missing pfpo-patch, all tests are passing.

pmur:
Thanks Stefan! I've updated the commit messages with a signoff. Otherwise they should be unchanged.

Issue #94, opened 2019-10-29
TS 18661 Support


pmur:
I would be nice to extend support for TS 18661-2.

This will require fixing up the headers to honor new feature tests, and implement new API. I see the following API which is missing:

nextdowndN
nextupdN
canonicalizedN
fmaxmagdN
fminmagdN
totalordermagdN
totalorderdN
ufromfpdN
fromfpdN
fromfpxdN
ufromfpxdN
setpayloadsigdN
setpayloaddN
getpayloaddN
encodedecdN
decodedecdN
encodebindN
decodebindN

Where N is 32, 64, and 128.

pmur:
Issue #154 addes strfromdN

PR #95, opened 2019-11-01
NextupN function

pdrocaldeira:
nextupN implementation of TS 18661-2.

It's a nextafter but always towards positive infinity

Related to #94


pmur:
I recommend squashing these into a single patch.

dfp/math.h
```diff
@@ -265,6 +265,15 @@ extern _Decimal32 __rintd32 (_Decimal32 __x) __THROW;
...
 extern _Decimal32 __nextafterd32 (_Decimal32 __x, _Decimal32 __y) __THROW __attribute__ ((__const__));
 
+extern _Decimal32 nextupd32 (_Decimal32 __x) __THROW __attribute__ ((__const__));
+extern _Decimal32 __nextupd32 (_Decimal32 __x) __THROW __attribute__ ((__const__));
+
+extern _Decimal64 nextupd64 (_Decimal64 __x) __THROW __attribute__ ((__const__));
+extern _Decimal64 __nextupd64 (_Decimal64 __x) __THROW __attribute__ ((__const__));
+
+extern _Decimal128 nextupd128 (_Decimal128 __x) __THROW __attribute__ ((__const__));
+extern _Decimal128 __nextupd128 (_Decimal128 __x) __THROW __attribute__ ((__const__));
```
pmur: The TS 18661-2 prototypes should be guarded by the definition of `__STDC_WANT_IEC_60559_DFP_EXT__`.

ieee754r/nextupd64.c
```diff
@@ -0,0 +1,29 @@
+/* Returns the next representable neighbor of x in the direction of y
+
+   Copyright (C) 2006 IBM Corporation.
+   Copyright (C) 2007-2015 Free Software Foundation, Inc.
```
pmur: This should be updated to 2019. I don't think the IBM copyright is relevant to newly added files.

ieee754r/nextupd64.c
```diff
@@ -0,0 +1,29 @@
...
+
+   This file is part of the Decimal Floating Point C Library.
+
+   Author(s): Joseph Kerian <jkerian@us.ibm.com>
```
pmur: Likewise.

tests/nextup.input
```diff
@@ -0,0 +1,19 @@
+# name nextup
+# arg1 decimal
+# ret  decimal
+
+0                                        0.000001 decimal32
+-0                                       0.000001 decimal32
+1                                        1.000001 decimal32
+2                                        2.000001 decimal32
+3                                        3.000001 decimal32
```
pmur: I would also add a test for the smallest possible negative value, and -0.

ieee754r/nextupd32.c
```diff
@@ -0,0 +1,58 @@
+/* Returns the next representable neighbor of x in the direction of plus infinity
+
+   Copyright (C) 2006 IBM Corporation.
+   Copyright (C) 2007-2015 Free Software Foundation, Inc.
```
pmur: The copyright and authors' should be updated.

ieee754r/nextupd128.c
```diff
@@ -0,0 +1,29 @@
+/* Returns the next representable neighbor of x in the direction of y
+
+   Copyright (C) 2006 IBM Corporation.
+   Copyright (C) 2007-2015 Free Software Foundation, Inc.
```
pmur: Author and copyright should also be fixed up here too.

tests/nextup.input
pmur: The existing test suite can also be used to substitute interesting values for each type (e.g infinite, min/max values).

ieee754r/nextupd128.c
pdrocaldeira: https://github.com/libdfp/libdfp/blob/21b9650b3c8015da74ef360cb86111864c08dc91/ieee754r/nextupd128.c#L7

pdrocaldeira: https://github.com/libdfp/libdfp/blob/21b9650b3c8015da74ef360cb86111864c08dc91/ieee754r/nextupd128.c#L3-L7

tests/nextup.input
pdrocaldeira: Now I'm using the test suite constants but I think some of them are wrong. For example, `DEC_MAX_EXP` is replaced by 90/369/6111 for 32/64/128 bits respectively and I think it should be 97/385/6145 based on ISO TR-24732.

There's negative 0 and smallest negative test for every type.

ieee754r/nextupd64.c
pdrocaldeira: https://github.com/libdfp/libdfp/blob/21b9650b3c8015da74ef360cb86111864c08dc91/ieee754r/nextupd64.c#L3-L7

ieee754r/nextupd64.c
pdrocaldeira: https://github.com/libdfp/libdfp/blob/21b9650b3c8015da74ef360cb86111864c08dc91/ieee754r/nextupd64.c#L3-L7

dfp/math.h
pdrocaldeira: Please give a look on this commit: https://github.com/libdfp/libdfp/pull/95/commits/88f0f10cf9f73578b901cdf52ce0f03a5066bb84

I've changed the Makefile to define this constant at compilation time. I think I'm cheating but I don't know how to do it in another way.

tests/nextup.input
pmur: The maximum exponent is dissimilar to BFP. The mantissa is always an integral value, thus the maximum exponent of the format is reduced depending on the precision of the format. The maximum representable exponent is a value with one significand leading the radix point e.g 1.000eX.

pdrocaldeira: Oh, now it makes sense. Thanks!

ieee754r/nextupd32.c
```diff
@@ -0,0 +1,57 @@
...
+}
+
+DEC_TYPE
+INTERNAL_FUNCTION_NAME (DEC_TYPE x)
+{
+	  DEC_TYPE z = IEEE_FUNCTION_NAME (x);
+	    if (!FUNC_D(__isfinite) (z) && FUNC_D(__isfinite) (x))
+		        DFP_ERRNO (ERANGE);
+	      return z;
+}
```
pmur: Errno should not be modified by nextup or nextdown.

pdrocaldeira: You mean, like that?
https://github.com/libdfp/libdfp/pull/95/commits/9b3d5c9d6b804cb033008ec107613f737345afb9

pmur:
Would it be possible to merge the nextup and nextdown into a single PR with 2 patches?

pdrocaldeira:
This code is now on #112

PR #96, opened 2019-11-06
NextdownN function

pdrocaldeira:
nextdownN implementation of TS 18661-2.

It's a nextafter but always towards negative infinity

Related to #94 and almost identical to #95 

This PR breaks `check-abi-libdfp` because of the new 3 functions.
Should I create a new version (1.0.15) for it to pass?


pdrocaldeira:
This code is now on #112

PR #97, opened 2019-11-07
fmaxmagN function

pdrocaldeira:
Implements fmaxmagdN from #94 as TS-18661 recommends.  

Tested on x86_64 and ppc64le.

From TS-18661:

>The fmaxmagfunctions determine the value of their argument whose magnitude is the maximum of  the  magnitudes  of  the  arguments:  the  value  of x if  |x|  >  |y|, y if  |x|  <  |y|,  and fmax(x,  y) otherwise. 

>If  just  one  argument  is  a  NaN,  the fmaxmagfunctions  return  the  other  argument  (if  both arguments are NaNs, the functions return a NaN).


ieee754r/fmaxmagd32.c
```diff
@@ -0,0 +1,61 @@
...
+
+  FUNC_CONVERT_TO_DN (&x, &dn_x);
+  FUNC_CONVERT_TO_DN (&y, &dn_y);
+
+  decContextDefault (&context, DEFAULT_CONTEXT);
+  decNumberMaxMag(&dn_result, &dn_x, &dn_y, &context);
+
+  FUNC_CONVERT_FROM_DN (&dn_result, &result, &context);
+
+  return result;
```
pmur: I think this should also implemented without explicitly falling back to decnumber.

This implementation can be moved into sysdeps/soft-dfp/

ieee754r/fmaxmagd32.c
```diff
@@ -0,0 +1,82 @@
...
+  printf("\n%lld\n", mantissaY);
+*/
+  if (diff > 0){
+    mantissaX *= pow(10,diff);
+  }
+  else {
+    diff *= -1;
+    mantissaY *= pow(10,diff);
+  }
+  if(mantissaX > mantissaY) return x;
```
pmur: why not something like `fabsdN(x) < fabsdN(y) ? y : x`?

This gives the compiler the freedom to choose the appropriate underlying implementation (e.g hard or soft). I am not sure if fabsdN is inlined today, but it is quick operation.

pdrocaldeira: But I think the `fabsdN(x)` uses libdecnumber. (Or I've just sabotaged myself for this whole week)
https://github.com/libdfp/libdfp/blob/af9e5e587b36872279ab7010e6f4ea583b590061/ieee754r/fabsd32.c#L40-L55

pmur: My suggestion is wrong when x == y. if the signs differ, I think the positive value should be returned.

pmur: That is the fallback version. However, both BID and DPD can trivially implement this by clearing the most significant bit. I believe PPC overrides these functions in sysdeps/powerpc/dfpu.

pdrocaldeira: Nevermind.  I've found out that we have versions for s390 and powerpc. :man_facepalming:

ieee754r/fmaxmagd32.c
```diff
@@ -0,0 +1,64 @@
...
+  if( FUNC_D(__isnan) (y) != 0) return x;
+ 
+  //Line below works for normal numbers and +/-Infinity 
+  if( FUNC_D(__fabs) (x) == FUNC_D(__fabs) (y))
+  {
+    if (x > 0) return x;
+    return y;
+  }
+  if( FUNC_D(__isinf) (x) != 0) return x; 
+  if( FUNC_D(__isinf) (y) != 0) return y;
```
pmur: This should be implicit in the comparison below.

pdrocaldeira: Done!
https://github.com/libdfp/libdfp/pull/97/commits/94f9fa0e77ee0eabbf74c18c230f8e959f8f9c27

pmur:
I think the squashed version looks OK. Can you squash this and test with x86?

configure.ac
```diff
@@ -1,7 +1,7 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.69])
-AC_INIT([libdfp],[1.0.14],[https://github.com/libdfp/libdfp/issues])
+AC_INIT([libdfp],[1.0.15],[https://github.com/libdfp/libdfp/issues])
```
pmur: Let's save this change for the formal commit which introduces version 1.0.15. Likewise, ./Versions.def should include this, I don't think I see it here.

Aside from some rebase conflicts, this looks OK.

sysdeps/s390/dfpu/libdfp-no-ti.abilist
```diff
@@ -306,6 +306,11 @@ LIBDFP_1.0.0
  wcstod64 F
 LIBDFP_1.0.13
  LIBDFP_1.0.13 A
+LIBDFP_1.0.15
```
pmur: This file no longer exists upstream. I would recommend a rebase

pmur:
I think this looks OK once you squash the fixup commit.

PR #102, opened 2019-11-08
Testing overhaul

pmur:
This is largely inspired by glibc, and very similar.

This will require regeneration of ulps on all targets as it is shaking up some very old dust.

This should only be committed once satisfactory workarounds are present. Sample ULPs can be observed for the soft targets. (e.g cos is broken around pi/2 today).


pmur:
It may also be helpful to extend the tester to verify non-default rounding modes don't generate wrong answers too.

pmur:
Folding this one into expd128.

Issue #103, opened 2019-11-21
ldexpd*(x,y) returns 10^(y+1)


pmur:
TS 18661-2 states it should be equivalent to x * 10 ^ y.

pmur:
Fixed in d67dc13e5bff82b5fe4b4a7ad590a7fb33ab6736

Issue #101, opened 2019-11-08
Tests cases generating warnings on 32b builds


pmur:
lrint.c and similar are rightfully generating warnings about integer overflow for several tests when built for ppc32 or i386. This was introduced by my large chunk of patches to cleanup the build system.

pmur:
Fixed

Issue #100, opened 2019-11-08
cos(1.57079697) fails


pmur:
This is generating a value with ~3 million ULPS, or more likely, a wrong value.

pmur:
Fixed

Issue #99, opened 2019-11-08
expd64(1000) fails 


pmur:
This value exceeds the log table lookup and suffers from undefined behavior.

pmur:
Note. this is only a bug in powerpc.

pmur:
Fixed in noted issues.

PR #105, opened 2019-11-22
Lots of changes leading up to implementing native expd128

pmur:
This is a meaty change to implement some useful functionality, especially for those with hardware dfp.

1. move expd64 to common code from ppc. This isn't PPC specific.
2. implement expd128 based on glibc float128 version
3. improve testing framework to verify the above does work
4. many related bug fixes along the way (e.g #22 #100 #99 #103 #102 #101) any maybe a few others I haven't mentioned

Tested with PPC64LE hard/soft, x86 soft.


pmur:
@stliibm if you have bandwidth, could you try these out? I expect the test suite to fail as the ULPs file will need to be regenerated (all going well the "make regen-ulps" command will work after make check fails)

stliibm:
@pmur I've build and test your patches. Yes you are right. The ulps needs to be adjusted.
I've pushed the commit "s390: update ulps" (https://github.com/stliibm/libdfp/commit/2e8f08501d2315fa863bb4eb728397f054bfe40d) to my branch expd128-on-s390.

"make regen-ulps" worked fine! Thanks.

Do you include my patch in your pull-request or shall I create a new pull-request after this one is merged?

pmur:
@stliibm let me push these, and you can signoff and create a PR. It LGTM.

PR #110, opened 2019-12-05
Generic fabs

### LINKED ISSUE #106: Implement fabsdN natively and generically
Both BID and DPD use the most significant bit to indicate sign. This can be done without the overhead of decnumber.

pdrocaldeira:
This function just set the MSB to 0.

Tested on x86_64 and ppc64. I ~didn't~ tested it on a big endian machine.

This closes #106

### DIFF
--- a/ieee754r/fabsd32.c
+++ b/ieee754r/fabsd32.c
@@ -1,11 +1,10 @@
 /* Calculate the absolute value for a _Decimal32 type
 
-   Copyright (C) 2006 IBM Corporation.
-   Copyright (C) 2007-2015 Free Software Foundation, Inc.
+   Copyright (C) 2007-2020 Free Software Foundation, Inc.
 
    This file is part of the Decimal Floating Point C Library.
 
-   Author(s): Joseph Kerian <jkerian@us.ibm.com>
+   Author(s): Joseph Kerian <jkerian@us.ibm.com>, Pedro Caldeira <caldeira@linux.ibm.com>
 
    The Decimal Floating Point C Library is free software; you can
    redistribute it and/or modify it under the terms of the GNU Lesser
@@ -28,8 +27,6 @@
 #  include <decimal32.h>
 #endif
 
-#include <decContext.h>
-#include <decNumber.h>
 #include <math.h>
 #include <ieee754r_private.h>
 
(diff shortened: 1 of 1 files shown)


ieee754r/fabsd32.c
```diff
@@ -40,19 +37,13 @@
...
-  decContextDefault (&context, DEFAULT_CONTEXT);
-  decNumberAbs (&dn_result, &dn_x, &context);
-
-  FUNC_CONVERT_FROM_DN (&dn_result, &result, &context);
-
-  return result;
+  DEC_TYPE result = x;
+  unsigned char *p = &result;
+  #if __BYTE_ORDER == __LITTLE_ENDIAN
+    p += _DECIMAL_SIZE/8 -1;
```
pmur: While not wrong, I would add a note and compiler check to ensure BID or DPD is used. Also minor nit, a space between - and 1.

pmur: Also, you could probably getaway with p += sizeof(DEC_TYPE) - 1

pdrocaldeira: >I would add a note and compiler check to ensure BID or DPD is used

How can I do that? :baby: 
I've saw only how to check about BID: `#ifdef __DECIMAL_BID_FORMAT__` 
Other thing is, if we are not using neither DPD or BID what is the desired behavior, stop compiling?

pmur: Ah, yes. I don't think there is an explicit test for DPD format.

ieee754r/fabsd32.c
```diff
@@ -1,11 +1,10 @@
 /* Calculate the absolute value for a _Decimal32 type
 
-   Copyright (C) 2006 IBM Corporation.
-   Copyright (C) 2007-2015 Free Software Foundation, Inc.
+   Copyright (C) 2007-2019 Free Software Foundation, Inc.
```
pmur: This should get updated to 2020. Otherwise, LGTM if you squash the changes.

pdrocaldeira: Done! https://github.com/libdfp/libdfp/blob/42de205d01ce4542cc3a4a94a6b439f7bdf43840/ieee754r/fabsd32.c#L3

PR #111, opened 2019-12-06
Add native logd128

pmur:
This is adapted from Stephen L. Moshier's ieee754 float128 provided
to glibc with decimal128 instead.

Lookup tables are generated, and the generator is provided intree
for verification and fixup if required. With some extra work, the
size could probably be reduced if we enforce rounding mode.

Likewise, a few subtle, yet important changes are made to better
ensure terms are summed from smallest to largest. Otherwise,
intermediate precision can be lost (e.g log(2)).


ieee754r/logd128.c
```diff
@@ -23,7 +24,120 @@
...
+  y =
+    ((((((((((((l15 * z + l14) * z + l13) * z + l12) * z + l11) * z +
+	    l10) * z + l9) * z + l8) * z + l7) * z + l6) * z + l5) * z +
+      l4) * z + l3) * z * w;
+  y -= 0.5DL * w;
+  y += e * ln10b;
+  y += z;
+  y += lntvals[m].lnt;
+
+  /* e*ln10a is always g.t 0. Subtract from this to avoid error on small values e.g x < ln(10) */
```
pmur: That should read g.t 1. In fact |e*ln10a| will be around the range 14000 - 2.3. Thus subtracting 1 is always exact.

Issue #121, opened 2020-01-07
make check fails with python3.


stliibm:
Generating the tests with python3 fails due to the following error:
File "tests/gen-libdfp-tests.py", line 281
op.cppchecks.append(cppcheckdict[earg])
TabError: inconsistent use of tabs and spaces in indentation

PR #122, opened 2020-01-07
Fix TabError in tests/gen-libdfp-tests.py. #121

stliibm:
Generating the tests with python3 fails due to the following error:
  File "tests/gen-libdfp-tests.py", line 281
    op.cppchecks.append(cppcheckdict[earg])
                                          ^
TabError: inconsistent use of tabs and spaces in indentation

This patch just exchanges the TAB with spaces.

Signed-off-by: Stefan Liebler <stli@linux.ibm.com>


pmur:
Thanks!

Issue #123, opened 2020-02-07
Cannot import dfp/math.h after math.h


arteymix:
Hello!

I am trying to fiddle your library into a small project to provide extended numerical types through the GLib type system (see https://github.com/arteymix/numeric-glib/pull/12). I also expose these types in such a way we can use Vala/Genie languages for operating at a higher level.

Vala generates code that imports `math.h` prior to `dfp/math.h` and there is an explicit check to prevent this import order in libdfp math header file.

What is the rationale behind this behaviour?

pmur:
This is due to a decision made many years ago not to integrate the _DecimalN features of ISO TR 24732 and later TS 18661-2 which extend the API of the standard C library and changes the behavior of many existing macro functions into glibc.  libdfp works around this by interposing C library headers which are modified by these proposals.  See Section 6 of README.user for more details.

In theory this works, in practice it has some warts.  Notably, the classification macro (e.g `isfinite` or `isinf`) are a bit outdated compared to their glibc equivalents, and scanf does not support scanning decimal types.

arteymix:
I see. The trick would be to include dfp include folder in C include path.

Have you considered providing a pkg-config file to automate this work for end-users?

arteymix:
Everything works on my side now.

PR #124, opened 2020-02-12
Add pkg-config file

arteymix:
As suggested in #123, this file will ease the usage of libdfp by setting a proper include directory for C headers and the expected `__STDC_WANT_DEC_FP__` definition.


arteymix:
It's effectively a trivial change. For what it's worth, I've never seen a copyright notice in a *.pc file.

Issue #125, opened 2020-02-18
fabs() macro is broken


pmur:
This is an almost type-generic macro (only with respect to _DecimalN) which will cause some weird headaches for binary float users.  This at best belongs in tgmath.h, and should be removed from math.h

pmur:
This is fixed with the aforementioned patch.

PR #128, opened 2020-02-20
Add _Float128 support, cleanup dfp/math.h, and move to C11

pmur:
This also fixes a number of little bugs along the way. Notably #125. Notably, ix86 also supports _Float128 with recent GCC, thus I've also removed the TFmode conditional from libbid so it actually builds.

I built test-bfp-conversions for hard-dfp and verified the macros evaluated TFmode correctly. I think it works, but should be verified. @stliibm if you have a moment, can you run tests on s390?


pmur:
@tuliom FYI, if you need to scratch that itch to review libdfp code.

stliibm:
@pmur I've just build and run the testsuite with this branch. Despite of the check-localplt issue #129 the testsuite runs fine.

pmur:
And tested for ppc{,64}.  Some workarounds were required to support the increasingly old RHEL7 4.8 toolchain which doesn't entirely support C11.  Will submit Monday if I don't hear anything.

PR #135, opened 2020-03-27
Improve configure

tuliom:
Recent changes to libdfp added new functions that weren't being correctly tested at configure time.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -522,28 +522,30 @@ AC_CHECK_FUNCS([floor gettimeofday memset nl_langinfo setlocale sqrt strchr strt
 gcc_min=4
 gcc_min_minor=6
 gcc_extra=""
-glibc_ver="2.10"
+glibc_min=2
+glibc_min_minor=10
 if test "$host_machine" == powerpc64le || test "$base_machine" == i386 || test "$host_machine" == x86_64; then
   gcc_min=6
   gcc_min_minor=2
-  gcc_extra="#if __GLIBC__ < 2 && __GLIBC_MINOR < 26
-             #error GLIBC 2.26 or newer required for this host
-             #endif"
-  glibc_ver="2.26"
+  glibc_min=2
+  glibc_min_minor=26
 fi
 
-# Copy the glibc check for GCC minimum version:
+# Based on the glibc check for GCC minimum version.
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -518,32 +518,42 @@ AC_TYPE_INT64_T
...
 
+# For _Float128 targets, we require yet newer glibc (2.26) and gcc (6.2)
+AS_IF([test "$host_machine" == powerpc64le || test "$base_machine" == i386 || test "$host_machine" == x86_64],[
+AC_CACHE_CHECK([if compiler/libraries support _Float128], libc_cv_float128_ok, [
+AC_TRY_COMPILE([#define __STDC_WANT_IEC_60559_TYPES_EXT__
+#include <math.h>], [
+_Float128
+foo(_Float128 a)
+{
+  return sinf128(a);
```
pmur: I don't agree this change improves the configury.  This probably fails on GCC 6.2 unless CFLAGS contains -mfloat128.  The build system takes care of that dependency.  We are testing for _Float128 and strfromf128 and strtof128 support in the C Library.  Note, there was a bug fixed by ef3942dd19699aeb4e3f666e4613548d905c7fc6 which failed to check ppc64le properly.

tuliom: > We are testing for _Float128 and strfromf128 and strtof128 support in the C Library.

I don't see anything like that at configure.ac.  What am I missing?
I see a code checking for the glibc version using `gcc_extra`, but it doesn't work because it does not include `features.h`.

>  This probably fails on GCC 6.2 unless CFLAGS contains -mfloat128.

Indeed.  I can fix that if you agree that testing for the feature is better than testing for the version.

pmur: features.h is more direct, using that directly is ok. My original line of thinking is that stdio.h must include it enable feature dependent macros (e.g str{to,from}f128) without having to drive GCC with arch specific options. 

Likewise, I'd prefer to check glibc/gcc version directly than indirectly test features which may require arch specific options.  The configure file is getting too complex.

tuliom: OK.  Let me change the patch.

tuliom:
@pmur I've made the modifications you suggested.

pmur:
Thanks, LGTM

PR #138, opened 2020-05-12
Relax submachine tests to support any power* CPU

pmur:
This reduces a bit of deprecated noise from sysdeps by removing the nofpu and fpu directories.  It is assumed all supported targets have hard-fp.  I haven't tested or tried soft-fp, ever, in 5 years.

This fully decouples and effectively deprecates --with-cpu.  It can optionally supply a CPU compiler option if none is passed via env, and similarly it can be used to disable hardware dfp if "no" is passed.

Notably, I have left the Z test alone.  It coerces a compiler into supporting hard-dfp if the `-mzarch` happens to be the minimal number of options to enable hardware dfp.  I think all modern compilers do this anyways.  It would be nice to defer to a single test, but I don't think this obstructs more generic submachine usage.


pmur:
@tuliom The title is wrong here now.  This will work for any CPU irrespective of name  assuming consistent GCC intrinsics (as both BID and DPD do today).

configure
```diff
@@ -5333,13 +5263,29 @@ EOF
...
+    *extendddtd*)
+      with_dfp=no
+      ;;
+    *)
+      with_dfp=yes
+      ;;
+  esac
+fi
+
 if test -n "$submachine"; then
```
rzinsly, line 5289: Will this ever be not null?

pmur: No, with hindsight, it should be set by --with-cpu to preserve the old behavior of setting -mcpu/-march.  Thanks.

PR #141, opened 2020-07-07
Use python3 when running bare scripts

pmur:
The Makefile will explicitly invoke the detected python interpreter,
however these scripts should use python3 if run directly by a
developer.


pmur:
It isn't strictly needed, but it is a cleanup.  Sometimes it is handy to run the scripts in isolation when debugging.

PR #145, opened 2020-07-24
Fix .pc permissions

frediz:
pkg config files are installed with executable rights which shouldn't
be.


pmur:
@frediz can you add a signoff? Otherwise, LGTM.

frediz:
Should be better. Thanks.

Issue #150, opened 2020-08-10
Issues when building with LTO


tuliom:
The Fedora community is enabling LTO on all its packages and noticed a couple of issues when running libdfp tests.
Notice that Stefan already investigated the s390x issues and confirmed they're caused by GCC issues.
Need to investigate what is causing the issues on ppc64le and x86_64.

Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1863988

Errors on ppc64le:

```
Details of failed tests:
cat ./check-localplt.out 
Extra PLT reference: libdfp.so: __dpd_extenddftd
Extra PLT reference: libdfp.so: __dpd_extenddfdd
Extra PLT reference: libdfp.so: __dpd_truncdfsd
Extra PLT reference: libdfp.so: __dpd_trunctddf
Extra PLT reference: libdfp.so: __dpd_trunctdsd2
cat ./test-acos-d128.out 
cat ./test-acos-d64.out 
cat ./test-fmaxmag-d128.out 
cat ./test-fmaxmag-d64.out 
cat ./test-fminmag-d128.out 
cat ./test-fminmag-d64.out 
FAIL: check-localplt
FAIL: test-acos-d128
FAIL: test-acos-d64
FAIL: test-fmaxmag-d128
FAIL: test-fmaxmag-d64
FAIL: test-fminmag-d128
FAIL: test-fminmag-d64
Summary of test results:
      7 FAIL
    161 PASS
```

Errors on x86_64:
```
Details of failed tests:
cat ./check-localplt.out 
Extra PLT reference: libdfp.so: __bid_floatsidd
Extra PLT reference: libdfp.so: __bid_eqdd2
Extra PLT reference: libdfp.so: __bid_unordtd2
Extra PLT reference: libdfp.so: __bid_addsd3
Extra PLT reference: libdfp.so: __bid_divdd3
Extra PLT reference: libdfp.so: __bid_unordsd2
Extra PLT reference: libdfp.so: __bid_floatunsdisd
Extra PLT reference: libdfp.so: __bid_floatditd
Extra PLT reference: libdfp.so: __bid_nedd2
Extra PLT reference: libdfp.so: __bid_floatunsdidd
Extra PLT reference: libdfp.so: __bid_gtdd2
Extra PLT reference: libdfp.so: __bid_subdd3
Extra PLT reference: libdfp.so: __bid_floatsitd
Extra PLT reference: libdfp.so: __bid_addtd3
[...]

PR #152, opened 2020-08-13
Ignore PLT references from GCC intrinsic functions

pmur:
When LTO is enabled, intrinsic function usage no longer
bypasses the PLT.  They aren't really C functions and
don't really exist to rename above the ASM level which is
ignored by LTO.

Instead, just ignore such references and generate noise on
stderr if they are found.  There isn't much we can do, and
the performance impact is likely marginal.

This fixes Issue #150 for some definitions of fix.

### DIFF
--- a/scripts/check-localplt.awk
+++ b/scripts/check-localplt.awk
@@ -57,6 +57,12 @@ NF != 2 && !(NF == 4 && ($3 == "RELA" || $3 == "REL")) {
       delete accept_type[key]
   } else if (key in accept) {
     delete accept[key]
+  } else if ($2 ~ /__dpd_.*/ || $2 ~ /__bid_.*/) {
+    # Intrinsic PLT references may not be avoidable.
+    # This happens if -flto is used, so record and make
+    # noise.  Most of these should have minimal performance
+    # impact anyways.
+    intrinsic_plt[key] = key
   } else {
     print "Extra PLT reference:", $0;
     if (result == 0)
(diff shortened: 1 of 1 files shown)


tuliom:
I wonder if this is a linker issue instead.
Interestingly, wcstod32 does get it right, but extendtftd does not.

The relocations do point to the expected symbol:
```
$ objdump -x ../libdfp-build/extendtftd.os
...
RELOCATION RECORDS FOR [.text]:
OFFSET           TYPE              VALUE
...
00000000000000fc R_PPC64_REL24     __GI___dpd_extenddftd
0000000000000110 R_PPC64_REL24     __GI___dpd_extenddftd
```

But it seems that, for this particular case, the linker prefers to use `__dpd_extenddftd` instead of `__GI___dpd_extenddftd`.

pmur:
@tuliom did you mean __dpd_extend**tf**td above? __dpd_extendtftd implicitly calls the DF -> TD on both doubles in the ibm128 value, and combines them.  That is one of the few places you will see internal PLT references on ppc64/hard-dfp where the compiler cannot inline them.

tuliom:
> did you mean __dpd_extendtftd above?

@pmur No.

>  __dpd_extendtftd implicitly calls the DF -> TD on both doubles in the ibm128 value, and combines them.

I agree with you that a call is needed. It should have been a local call, but the linker is replacing the local symbol with a call through the PLT.
I still do not understand why this is not linker bug.
I do understand asm statements are lost. I do not understand why it's correct to replace a local symbol with a global one.

tuliom (approved):
I now understand the output from objdump was actually coming from a fat LTO object and is not useful for the analysis here.

I still believe we have to find another solution to this, but there are no reasons to continue blocking this patch.
I need to study and I need more time for that.

With that said, I'm approving this patch and merging it.

Thanks for your patience!

PR #153, opened 2020-09-01
Enable travis CI for libdfp.

stliibm:
According to https://travis-ci.com/plans, travis-CI is "Free for open source".
Thus this implements a small .travis.yml file which adds builds for amd64, ppc64le
and s390x. The latter two also have a version for software dfp.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+dist: focal
+language: c
+
+matrix:
+  include:
+    - name: "linux-amd64-gcc"
+      os: linux
+      arch: amd64
+      compiler: gcc
+      env: CFLAGS="-O2"
+    - name: "linux-ppc64le-gcc"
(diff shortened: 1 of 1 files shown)


pmur:
I think this looks OK.  Can it support running i386 or s390 tests too? Slightly related, but are there any opinions on deprecate the <64b targets?

stliibm:
I've force pushed the commit. Is there a better way to update the commit-message?
Unfortunately there are only those archs:
https://config.travis-ci.com/ref/arch
- amd64
- arm64
- ppc64le
- s390x
Thus I think I have to keep my local s390(31bit) test.

tuliom:
maybe, we could use the same infrastructure and cross compile for the other architectures.
It wouldn't be able to run the tests, but it could at least spot some build issues before patches are merged.

PR #154, opened 2020-09-02
Add TS18661-2 strfromdN

pmur:
This is a creative wrapper around the libdfp printf hook.

Fortunately, the behavior is a restricted subset of the
printf family of IO, thus keeping life simple.  The only
gotcha is that using memory files requires dynamic allocation
which could fail.

test-printf is enhanced to test the strfromd* variants under
nominal circumstances (i.e valid format and unlimited buffer).

test-strfrom is added to capture some more exotic usage of
strfrom.  That is, length computations and bogus format
strings.

Likewise, fix bug in scaffold.c which uses the wrong array
size to buffer. CHAR_MAX is both misleading, and too small
on systems where it is signed for the existing test cases.

Signed-off-by: Paul E. Murphy <murphyp@linux.vnet.ibm.com>


stdlib/strfromd32.c
```diff
@@ -0,0 +1,90 @@
...
+
+  /* TODO: this is probably better than crashing.  The standard isn't so clear
+     about this.  */
+  if (!f)
+    {
+    /* errno set by fmemopen.  */
+    return 0;
+    }
+
+  info.user = sizeof (fp) == 4 ? mod_H : (sizeof (fp) == 8 ? mod_D : mod_DD);
```
pmur, line 58: This may not work correctly if glibc registration fails, or a static library user does not run the onetime setup.  Possible workaround is to statically assign unique 2^n values for mod_{H,D,DD} in case registration fails or is not used.

PR #159, opened 2021-01-29
Run tests on gcc-9 and gcc-10

tuliom:
We've just seen an issue happening on gcc-11.
This is a good change to start testing on different GCC versions.
Unfortunately, gcc-11 is not yet available on Ubuntu [1].

[1] https://launchpad.net/ubuntu/+source/gcc-11

Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,31 +3,105 @@ language: c
 
 matrix:
   include:
-    - name: "linux-amd64-gcc"
+    - name: "linux-amd64-gcc-9"
       os: linux
       arch: amd64
-      compiler: gcc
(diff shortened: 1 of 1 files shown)


pmur:
@tuliom  I think we need to add g++ too.  There are a few tests against C++ headers we supplied by libdfp.

tuliom:
Indeed.  I'll rework this.
Thanks!

tuliom:
@pmur The patch has just been updated to force usage of g++ versions too.

Issue #164, opened 2021-10-02
Improper behavior of operator>>


dacodas:
I don't believe that the [`std::basic_istream<charT, traits> & operator>>(std::basic_istream<charT, traits> & is, decimal32 & d)`][1] definitions adhere to the C++ [FormattedInputFunction][2] named requirement.

[There][3] [is][4] [a][5] [whole][6] lot happening in the GCC machinery that implements this.

The main symptom I noticed is that when this library calls `.read()` on the `istream` during `operator>>`, it eats through `CHAR_MAX` `char`s of the input stream, which is definitely unexpected.

The expected behavior is described in [cppreference.com's page for std::num_get][7].

> In any case, the check is made whether the char obtained from the previous steps is allowed in the input field that would be parsed by std::scanf given the conversion specifier selected in Stage 1. If it is allowed, it is accumulated in a temporary buffer and Stage 2 repeats. If it is not allowed, Stage 2 terminates. 

Rather than consuming `CHAR_MAX` `char`s, these `operator>>` functions should only consume `char`s until it finds a `char` that wouldn't be accepted by the conversion. 

I will attempt in the next few days to submit a pull request that addresses this.

[1]: https://github.com/libdfp/libdfp/blob/308a2762d46204d448ee2a2c50d52bccb93e3af9/dfp/decimal/decimal#L213
[2]: https://en.cppreference.com/w/cpp/named_req/FormattedInputFunction
[3]: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/std/istream;h=5ad60dbd709f575c9e9fc9ecf8eafec1b91ec9fc;hb=HEAD#l218
[4]: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/bits/istream.tcc;h=2a153c2e140a5267b3e67d00cdea3450a2c9d7bb;hb=HEAD#l91
[5]: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/bits/locale_facets.tcc;h=310bfed65c4318be52c8d48fa17e29a593e9a69d;hb=HEAD#l703
[6]: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/bits/locale_facets.tcc;h=310bfed65c4318be52c8d48fa17e29a593e9a69d;hb=HEAD#l145
[...]

PR #166, opened 2021-10-03
operator>> semantics

dacodas:
This pull request attempts to address #164 
There are still a few TODOs

- [ ] Add as comments where the code is calqued from
- [ ] Make sure formatting matches libdfp project
- [ ] Currently, localized floats don't work correctly


dacodas:
Alright, I have written some tests, and added this operator for the other types. I have also signed off on my squashed commit.

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+      string __xtrc;
+      __xtrc.reserve(32);
+      __beg = this->_M_extract_float(__beg, __end, __io, __err, __xtrc);
+      std::__convert_to_v(__xtrc.c_str(), __v, __err, locale::facet::_S_get_c_locale());
+      if (__beg == __end)
+        __err |= ios_base::eofbit;
+      return __beg;
+    }
+
+  template<typename _CharT, typename _InIter>
```
pmur, line 392: Is it possible to parameterize the decimal type here to avoid duplicating this for each decimal type?

dfp/decimal/decimal
```diff
@@ -209,38 +489,42 @@ namespace decimal
...
-	return is;
+        static_assert(
+            std::is_same<T, decimal::decimal32>::value
+              || std::is_same<T, decimal::decimal64>::value
+              || std::is_same<T, decimal::decimal128>::value,
+            "This template function is only meant to be used with "
+            "decimal floating point types"
+        );
+
+        auto& dfpIstream {
```
pmur, line 519: Will this only work with C++11 and newer?

tests/test-istream.cpp
```diff
@@ -51,43 +53,47 @@ typedef struct{
...
-  {__LINE__, -9.999E-3DD, "-9.999E-03", "%Da", -1.0DD},
-  {__LINE__, -9.999E-3DD, "-0.009999", "%Da", -1.0DD},
-  {__LINE__, __builtin_infd64(), "inf", "%Da", -1.0DD},
-  {__LINE__, __builtin_infd64(), "INF", "%Da", -1.0DD},
-  {__LINE__, 4E384DD, "4.0E384", "%Da", -1.0DD},
+  {__LINE__, -9.999E-3DD, "-0.009999", 9, "%Da", -1.0DD},
+  {__LINE__, -9.999E-3DD, "-9.999000e-03", 13, "%Da", -1.0DD},
+  {__LINE__, -9.999E-3DD, "-9.999E-03", 10, "%Da", -1.0DD},
+  {__LINE__, -9.999E-3DD, "-0.009999", 9, "%Da", -1.0DD},
+  /* {__LINE__, __builtin_infd64(), "inf", "%Da", -1.0DD}, */
```
pmur, line 88: Should these two cases be commented out?

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+    }
+  }
+
+  template<>
+    void
+    __convert_to_v(const char* __s, decimal::decimal32& __v, ios_base::iostate& __err,
+		   const __c_locale&) throw()
+    {
+      // Assumes __s formatted for "C" locale.
+      const char* __sav = __set_C_locale();
```
pmur, line 216: Does this need to respect a locale? Or is that handled by the c++ stdlib?

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+	  __err = ios_base::failbit;
+	}
+
+      setlocale(LC_ALL, __sav);
+      delete [] __sav;
+    }
+
+  template<>
+    void
+    __convert_to_v(const char* __s, decimal::decimal128& __v,
```
pmur, line 290: Could these be parameterized based on the decimal type?

pmur (commented):
Sorry for the slowness at reviewing. How is locale handled in the stream class?

Likewise, if this is only C++11 or newer, there should be appropriate guard macros placed to protect users of old standards.

Thanks!

pmur:
@dacodas sorry for the delay.  I think it is OK to follow the layout of the existing code w.r.t parameterization. I was curious if it would be practical simplify them further. Anyhow, that can wait for another patch. Are you still working on this PR?

dacodas:
@pmur, no, no by no means. I wrote the above tasks on my own end to make sure I addressed them.

The ball was/is definitely in my court.

Let me take a look at pausing the PR until I can dedicate a bit more time towards making it happen.

dacodas:
Alright, for the time being I have marked this as a draft. I think in the next few weeks I can make those two small changes and re-submit.

Thank you for this great library, and looking forward to resubmitting in the near future!

dfp/decimal/decimal
jwakely: These are virtual functions, which cannot be templates.

dfp/decimal/decimal
jwakely: These functions *are already* parameterized on the decimal type. But they are specialied for each  of `float`, `double` and `long double`, and that's because they have to do different things, using `strtof`, `strtod` or `strtold` as appropriate. You could have a more generic function which then uses some *other* function (or more likely, a traits class) to do the non-generic pieces that need to differ for each decimal type. For example, this one calls `strtod128` and uses `DEC128_MAX`, the specializations for `decimal32` and `decimal64` do not.

In the GCC code the differences are more significant, e.g. `strtof` might not be available, and `strtold` might be available but broken (on mingw, IIRC), so using a single generic function that handles all three types is harder. Here, that's not true. The functions are almost identical except for whether they call strtod32, strtod64, or strtod128, and which constants they use, so the variation could easily be hoisted out into a traits class. Those constants should come from `std::numeric_limits` anyway, reducing the differences between the specializations even further. For some reason `numeric_limits` isn't specialized in GCC's `<decimal/decimal>` header, but almost certainly should be.

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+    struct _Save_errno
+    {
+      _Save_errno() : _M_errno(errno) { errno = 0; }
+      ~_Save_errno() { if (errno == 0) errno = _M_errno; }
+      int _M_errno;
+    };
+
+    // calls setlocale(LC_ALL, "C") and returns a string containing the old
+    // locale name. Caller must delete[] the string. Returns NULL on error.
+    const char*
```
jwakely, line 195: If this code is C++11-only, then this should return `std::unique_ptr<char[]>` and not require the caller to `delete[]` it. That's not an option in the original GCC code because it has to support C++98 still.

Blindly copying the GCC code is not a good idea, there are a number of constraints that apply to GCC which do not apply here. If we didn't need to support C++98 the GCC code would look different.

tests/test-istream.cpp
```diff
@@ -39,6 +39,8 @@
 
 #include <math.h>
 
+template class std::num_get<char, std::istreambuf_iterator<char, std::char_traits<char>>>;
```
jwakely, line 42: Is this explicit instantiation necessary? Why?

dfp/decimal/decimal
jwakely: `is_same` is trivial to reimplement in C++98.

dfp/decimal/decimal
```diff
@@ -209,38 +489,42 @@ namespace decimal
...
+            "decimal floating point types"
+        );
+
+        auto& dfpIstream {
+          static_cast<
+            dfp_basic_istream<charT, traits>&
+          >(is)
+        };
+
+        dfpIstream._M_extract(d);
```
jwakely, line 525: Isn't this undefined behaviour if this function is ever called with an istream that isn't the base-class subobject of a `dfp_basic_istream`? Shouldn't it at least use `dynamic_cast`?

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+      __xtrc.reserve(32);
+      __beg = this->_M_extract_float(__beg, __end, __io, __err, __xtrc);
+      std::__convert_to_v(__xtrc.c_str(), __v, __err, locale::facet::_S_get_c_locale());
+      if (__beg == __end)
+        __err |= ios_base::eofbit;
+      return __beg;
+    }
+
+  template <class charT, class traits = char_traits<charT>>
+  class dfp_basic_istream;
```
jwakely, line 408: Why add new istream types? That defeats the entire purpose of using istreams, which is that all istream types provide the same API and you don't need to care whether you're using `ifstream` or `istringstream` etc.

If users can only extract decimal types when using `dfp_basic_istream` then how do they interact with `ifstream` etc?

 This is **not** how you extend iostreams for custom types.

Also, `basic_dfp_istream` would be more consistent with naming conventions. But this type should just not exist at all.

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+      {
+        typename basic_istream<_CharT, _Traits>::sentry __cerb(*this, false);
+        if (__cerb)
+          {
+            ios_base::iostate __err = ios_base::goodbit;
+            __try
+              {
+                locale dfp_locale {this->_M_ios_locale, new dfp_num_get<_CharT>()};
+                auto& facet {use_facet<dfp_num_get<_CharT>>(dfp_locale)};
+                facet.get(*this, 0, *this, __err, __v);
```
jwakely, line 448: Why bother to create a locale using the facet, and then extract the facet again (with expensive RTTI checks to make sure it's present and has the right type)? You just need the facet.

```c++
dfp_num_get<_CharT> ng;
ng.get(*this, 0, *this, __err, __v);
```

dfp/decimal/decimal
```diff
@@ -180,6 +180,286 @@ public:
...
+	}
+
+      setlocale(LC_ALL, __sav);
+      delete [] __sav;
+    }
+
+
+
+  template<typename _CharT, typename _InIter = std::istreambuf_iterator<_CharT>>
+  class dfp_num_get : public num_get<_CharT, _InIter> {
```
jwakely, line 329: I don't understand the naming scheme in this code. You're using reserved names like `__save_c_locale()` as though you're part of the C++ implementation (which I suppose this library is intended to be, as an implementation of an ISO TS). But then you're just adding new names like `dfp_num_get` and `dfp_basic_istream` which are not in the implementer's namespace, and so would clash with user macros using those names.

Either stop using `__reserved` names for parameters, local variables etc. or use them for everything, and don't add new non-reserved names that are not part of the C++ standard or the decimal types TR. The draft I have of TR 24733 calls this `extended_num_get`.

dacodas:
@jwakely, thank you so much for your review and comments—I value them greatly

I'm going to shake off the cobwebs surrounding my thinking on this, internalize your comments, and incorporate your suggestions to the MR. It may take some time

Issue #169, opened 2021-11-10
Memory leak in `strtod32` etc.


ian-abbott:
Eyeballing `FUNCTION_L_INTERNAL()` in "stdlib/strtod32.c", not all paths through the function call `freelocale(C_locale);` before returning.

pmur:
This should be fixed with 437c154655625c408316a18767727626e1cbe9e4

PR #171, opened 2021-11-12
stdlib: Fix locale memory leak in strtodN/wcstodN

ian-abbott:
These all call an internal function named by the `FUNCTION_L_INTERNAL`
macro in "stdlib/strtod32.c" which calls `newlocale` to allocate a
locale object referred to by the variable `C_locale`.  Only 7 out of 13
return statements have a preceding call to `freelocale(C_locale)`,
leading to a likely memory leak if the program has called `setlocale`.
Always call `freelocale(C_locale)` before returning to plug the leak.

Signed-off-by: Ian Abbott <abbotti@mev.co.uk>

### DIFF
--- a/stdlib/strtod32.c
+++ b/stdlib/strtod32.c
@@ -512,6 +512,7 @@ FUNCTION_L_INTERNAL (const STRING_TYPE * nptr, STRING_TYPE ** endptr,
 	}
 
       /* It is really a text we do not recognize.  */
+      freelocale(C_locale);
       RETURN (0.0, nptr);
     }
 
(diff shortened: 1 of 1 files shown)


ian-abbott:
Fixes #169

pmur (approved):
Thanks, LGTM. The only trivial nit is the first word after the colon should be lower-case in the commit title.

tuliom (approved):

pmur:
I fixed the trivial capitalization in the commit title, and pushed it via CLI. Thanks @ian-abbott.

Issue #170, opened 2021-11-12
Dodgy hexadecimal parsing in `strtod32` etc.


ian-abbott:
Looking at `FUNCTION_L_INTERNAL()` in "stdlib/strtod32.c":

1. The `TOLOWER()` calls in the following extract seem to be in the wrong place:

```
  /* If no other digit but a '0' is found the result is 0.0.
     Return current read pointer.  */
  if ((c < L_('0') || c > L_('9'))
      && (base == 16 && (c < (CHAR_TYPE) TOLOWER (L_('a'))
			 || c > (CHAR_TYPE) TOLOWER (L_('f'))))
```

It does not allow `c` to be upper-case `'A'` through `'F'`.

2. In the following extract, it seems to be incorrectly treating `'g'` and `'h'` as hexadecimal digits:

```
	  /* There might be thousands separators or radix characters in
	    the string.  But these all can be ignored because we know the
	    format of the number is correct and we have an exact number
	    of characters to read.  */
#ifdef USE_WIDE_CHAR
	  if (*startp < L_('0') || *startp > L_('9'))
	    if (base==10 || (*startp < L_('a') || *startp > L_('h')))
	      ++startp;
#else
	  if (*startp < L_('0') || *startp > L_('9'))
	    if (base==10 || (*startp < L_('a') || *startp > L_('h')))
```

pmur:
Would you have some representative test cases to demonstrate? I suspect this code was inherited and adopted from glibc. Looking at the test coverage, I am unsure if base 16 inputs have been considered.

ian-abbott:
Here are some test cases. The handling of the fractional part after the decimal point is also incorrect.

```
  {__LINE__, "0xabcdef", 11259370.000000DF, 11259375.000000DD, 11259375.000000DL},
  {__LINE__, "0XABCDEF", 11259370.000000DF, 11259375.000000DD, 11259375.000000DL},
  {__LINE__, "0xabcd.e", 43981.875000DF, 43981.875000DD, 43981.875000DL},
  {__LINE__, "0XABCE.E", 43981.875000DF, 43981.875000DD, 43981.875000DL},
 ```

pmur:
Refreshing my knowledge of strtodN as standardized by TS-18661 part 2, only strings expressed in base-10 are acceptable and parsed.  strto{f,d,ld,fN} (from C99 and TS-18661 part 3) convert base 10 and base 16.

[...]

PR #172, opened 2021-11-16
stdlib: Restrict strtodN/wcstodN to parsing base-10 only

ian-abbott:
Quoting Paul Murphy in
https://github.com/libdfp/libdfp/issues/170#issuecomment-967476990 :
> Refreshing my knowledge of strtodN as standardized by TS-18661 part 2,
> only strings expressed in base-10 are acceptable and parsed.

The strtodN/wcstodN functions currently have some code to parse floating
point numbers in hexadecimal form, but it is quite buggy; it only
handles integers with lower case hexadecimal digits correctly.  Upper
case digits, fractional parts, and binary exponents are not handled
correctly, and values are truncated towards zero rather than rounded.

Just remove the code to handle hexadecimal floating point numbers.  A
hexadecimal floating point number will be converted to +0 or -0, and the
end pointer will point to the 'x' or 'X' that follows the leading '0'.

Signed-off-by: Ian Abbott <abbotti@mev.co.uk>

### DIFF
--- a/stdlib/strtod32.c
+++ b/stdlib/strtod32.c
@@ -90,7 +90,6 @@ extern unsigned long long int ____wcstoull_l_internal (const wchar_t *, wchar_t 
 # define L_(Ch) L##Ch
 # define ISSPACE(Ch) iswspace_l ((Ch), loc)
 # define ISDIGIT(Ch) iswdigit_l ((Ch), loc)
-# define ISXDIGIT(Ch) iswxdigit_l ((Ch), loc)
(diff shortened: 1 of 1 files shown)


pmur (commented):
This LGTM with a couple of minor change requests:

1. The test cases should be updated to verify this.
2. Lower-case "Restrict" in the commit title.
3. The test suite should be verifying the remainder of the string return from strtodN, notably for these cases where a hex string is passed. I can make these.

PR #177, opened 2022-01-26
tests: stop assuming long double type in test-bfp-conversions

tuliom:
Prepare for when distros switch to IEEE 128-bit floating point long
double on powerpc64le.


pmur:
Unfortunately, this highlights some other subtle issues inside libdfp. All the old TFmode functions are being compiled incorrectly. They assume an ibm128 long double.

Either the configury needs to pass -mabi=ibmlongdouble when building on ppc64le and ldouble is ieee128, or the internals need to be more specific about what format they should be using. I like the latter better.

pmur:
I've updated my commit to ensure we the libgcc __ibm128 functions are correctly compiled, and continue being tested with the abi switch.

PR #179, opened 2022-09-13
Enable support for AArch64 (using BID format)

christophe-lyon:
I have recently added DFP support for AArch64 in GCC, and I used libdfp as testsuite.  This required small changes:

* update of config/dfp.m4 to recognize aarch64, and regeneration of libdecnumber/configure

* introduction of #ifdef __LIBGCC_HAS_XF_MODE__ in libbid/*xf*.c like what is now in GCC gcc/libgcc/config/libbid.  This is necessary because AArch64 does not support XFmode.

* in libbid/bid_gcc_intrinsics.h, define TFtype to long double on AArch64, which does not support __float128.  This is not needed in GCC's libbid because BID_HAS_TF_MODE is not defined in that case.

* creation of sysdeps/aarch64/

I am not sure if the discrepancy between this libbid and GCC's copy should be fixed, nor how.

To make the build succeed, I also had to update config.guess and config.sub to more recent versions, but I do not include that change in this patch because it is not strictly related to AArch64 enablement.

'make check' reports 172 PASS.


pmur:
I'll resync libbid with the latest gcc changes. It's grown a few more workarounds in libdfp than I recalled when I made my earlier comment.

christophe-lyon:
> Hi @christophe-lyon, thank you for your contribution to libdfp. I see you've also done the enablement work on libgcc too. I wonder if we want to synchronize the other changes you've made to libbid too.
> 
> Is BID support turned on for aarch64 in a released gcc toolchain?
> 
> For my due diligence, have you read the CONTRIBUTING file, and have FSF copyright assignment or can agree the DCO?

Hi! Thanks for the quick answer :-)
Yes I have read the CONTRIBUTING file, everything is OK, I am the author of the patch. Sorry I missed the Signed-off-by tag, because it's not customary in GCC dev... 

I work for Arm, we do have FSF copyright assignment in place.

Thanks for doing the needful regarding libbid.

christophe-lyon:
I've noticed that the CI fails on amd64, it that really related to my patch?
Does Travis provide AArch64 (aka arm64) hosts, which you'll need to properly exercise the patch? (But as mentioned earlier, you also have to update config.sub/config.guess, which I prefer you do rather than me as it's not directly related to aarch64 and might cause further breakages I wouldn't be able to investigate).
Thanks

pmur:
> I've noticed that the CI fails on amd64, it that really related to my patch?

Yes, a minor change is needed in the bid configuration for x86. The libgcc copy of those functions are getting linked in, they don't support rounding modes, and don't have the hacks to expose the ABI. Those are causing the odd failures on x86.


> Does Travis provide AArch64 (aka arm64) hosts, which you'll need to properly exercise the patch?

It seems like they provide two options. If you want to try adding it, I found some notes at https://docs.travis-ci.com/user/multi-cpu-architectures . If not, I'll try. Is there a minimum GCC version where arm64 supports dfp?


> I work for Arm, we do have FSF copyright assignment in place.

Thanks. The DCO is optional since you've already assigned copyright. The DCO hopefully makes it easier for those who don't.

pmur:
I've updated libbid here. Hopefully the only fix you need to carry over is to disable TFmode functions for arm64.

christophe-lyon:
> > I've noticed that the CI fails on amd64, it that really related to my patch?
> 
> Yes, a minor change is needed in the bid configuration for x86. The libgcc copy of those functions are getting linked in, they don't support rounding modes, and don't have the hacks to expose the ABI. Those are causing the odd failures on x86.
> 
Strange, I remember building & checking on x86 with my changes....

> > Does Travis provide AArch64 (aka arm64) hosts, which you'll need to properly exercise the patch?
> 
> It seems like they provide two options. If you want to try adding it, I found some notes at https://docs.travis-ci.com/user/multi-cpu-architectures . If not, I'll try. Is there a minimum GCC version where arm64 supports dfp?
> 
I don't think I'll have time to investigate this, sorry.
GCC support is only in trunk for the moment, it will be in the next release.... next year :-(

christophe-lyon:
> I've updated libbid here. Hopefully the only fix you need to carry over is to disable TFmode functions for arm64.

You mean I have to rebase my patch over your updated trunk and resubmit?

pmur:
> You mean I have to rebase my patch over your updated trunk and resubmit?

Yes, it should be a trivial rebase. I also regenerated the configury and verified it stays the same with your changes.

pmur:
Thanks. I started to try cross-compiling this and found config.sub/config.guess also need updated. I'll update those too.

sysdeps/aarch64/Versions
```diff
@@ -0,0 +1,20 @@
...
+    __bid_extendsdtf;
+    __bid_extendddtf;
+    __bid_trunctdtf;
+  
+    __bid_trunctfsd;
+    __bid_trunctfdd;
+    __bid_extendtftd;
+  }
+
+  LIBDFP_1.0.0 {
```
pmur, line 12: I think these XFmode functions should be removed from aarch64. I am surprised this didn't cause the ABI test to fail.

christophe-lyon: Hmmm so I removed this second chunk, and re-ran 'make check' and I still get "172 PASS" and no error.
Then I re-added this chunk (so back to what this patch contains), still get "172 PASS".
Then I removed the 1st chunk, 'make check' still passed with "172 PASS".
Do I need to run 'make clean' or similar before running 'make check' ?
(ie. I somehow tried to deliberately break the checks and they still passed :-) )

christophe-lyon:
> Thanks. I started to try cross-compiling this and found config.sub/config.guess also need updated. I'll update those too.
Indeed, that's what I meant in the commit message.

sysdeps/aarch64/Versions
pmur: The version file gets merged into a version map file (libdfp.map) used by the linker.

ld doesn't seem to care if a symbol is listed in a version file, but not present in the link unit. The ABI check only inspects the symbols exported by shared library. Thus, these are quietly ignored.

Anyhow, it's best to remove these since they should not be provided by the aarch64 version at this time (likely never).

pmur:
Thanks. This LGTM.

christophe-lyon:
Thanks!

PR #195, opened 2024-01-31
ieee754r: fix powd64/32, add tests for it, and regen ulps

### LINKED ISSUE #188: MATH: POW: fail with powd64( 2, -1279 ) -> 0
can't say if bug, intentional, mistake by me, missed reading the manual?  ...  
managed to implement using of math functions in an application and in 'pure-C' code,  
experience the result '0' for the calculation in the subject,  
tried to put it in a small testprog:  
```  
        volatile _Decimal64 x1DD = 2.0DD; 
        volatile _Decimal64 x2DD = -1278.0DD; 
        printf( "powd64( 2.0DD, -1278.0DD ) : %.17DE \n", powd64( x1DD, x2DD ) ); 
        printf( "exp2d64( -1278.0DD )       : %.17DE \n", exp2d64( x2DD ) ); 
        x2DD = -1279.0DD; 
        printf( "powd64( 2.0DD, -1279.0DD ) : %.17DE \n", powd64( x1DD, x2DD ) ); 
        printf( "exp2d64( -1279.0DD )       : %.17DE \n", exp2d64( x2DD ) ); 
```  
output:  
```  
powd64( 2.0DD, -1278.0DD ) : 1.92161128896030000E-385 
exp2d64( -1278.0DD )       : 1.92161128896030000E-385 
powd64( 2.0DD, -1279.0DD ) : 0.00000000000000000E+00 
exp2d64( -1279.0DD )       : 9.60805644480100000E-386 
```  
_assume!_  fail based on 1278 being the max. **positive** exponent,  
but negative should reach out to -1324 ( quasi denormals ).  
_assume!_  similar fails for other powers ...

pmur:
Issue #188 uncovered a case where powd64 rounds to 0 incorrectly for small results. This also highlighted a lack of testing of pow.

Looking through the implementation highlighted several other issues.

Update gen-auto-dfp-tests to test the special cases defined by IEEE754 pow. Exception correction is not tested.

Likewise, use powd64 as powd32 on all platforms, not just ppc.

Fixes #188


stliibm:
Thanks for the hint.
I've pushed an ulps update here:
https://github.com/stliibm/libdfp/commit/66be5e448bc441bf71f39ce8ff579106c98d5f1b

Do you just pick up this commit? Or shall I create a new pull-request?

pmur:
Thanks, I'll pick it up when I'm ready to commit this.

newbie-02:
**If** ... I managed to implement the fix version successfully,  
and **if** ... Wolfram|alpha is a qualified reference,  
and **if** I didn't introduce other errors ...  
some results lack some accuracy.  
```
powd64( 3.19034302541825200E+13, -1.55493546491461400E+00 ) : 1.00555155985385600E-21 
Wolfram|alpha result :                                        1.005551559853848356269012840... × 10^-21 
powd64( 1.08507545139287500E+04, -1.98586216955218000E+00 ) : 9.68573922386492800E-09 
Wolfram|alpha result :                                        9.685739223864927096824376046... × 10^-9 
powd64( 1.96877857924400000E+09, -5.38711002611281000E-01 ) : 9.84269171597413800E-06 
Wolfram|alpha result :                                        9.842691715974125463736414499... × 10^-6 
powd64( 4.98084050185682800E+00,  8.31505582817438000E-01 ) : 3.80024007789110900E+00 
Wolfram|alpha result :                                        3.800240077878302582784863067... 
powd64( 2.64943091197714100E+09,  6.31392470927920100E-01 ) : 8.90651089453718200E+05 
Wolfram|alpha result :                                        890651.089453720374200784857... 
powd64( 3.41000000000000000E+01,  1.95000000000000000E+01 ) : 7.73945519450458300E+29 
Wolfram|alpha result :                                        7.73945519450458710555628329... × 10^29 
```

pmur:
I am not surprised. For exponents 0<y<1, it is computing `expd64(y*logd64(x))`. That isn't great for worst-case accuracy. It also raises a question of the accuracy of expd64/logd64 too.

This MR is only addressing the defects in the current implementation.  Improving worst case accuracy likely requires a careful rewrite. I'll add any test cases which demonstrate error. I measured 28833 ULPs with those inputs on powd64. That's less than ideal.

newbie-02:
< May be outdated by your _Decimal128 bandage, could offer better performance > 
  
This is an indecent proposal ( for a decimal-oriented data type ),  
but until we have something better:  
_Decimal64 -> long double conversion has very little 'approximation loss'.  
'powl' is not perfect but much better than the current powd64.  
`(_Decimal64)powl( (long double)x, (long double)y )`  
gave better results for the cases mentioned. I think it could be used as a substitute calculation for the time being and later to check original decimal calculations. 
  
( It is a bit of a pity for a data type with the potential and the good intention for 'decimal correct' to spread such inaccuracies after such long time. And no wonder if it is correspondingly rarely used. )

pmur:
Note, radix conversions are not free and will introduce rounding errors. Also, long double does not map to a consistent format on all platforms (or within a platform, see ppc64le) so the results will vary. expd128 and logd128 are native decimal, and based off the implementations in glibc for float128. They should be fairly accurate.

I am trying to avoid regressing back to the libdecnumber soft-dfp pow function. If you find a value with more than 2 digits of inaccuracy with my workaround, I will revert to libdecnumber pow (which is correct, but much slower).

newbie-02:
Looks much better now, can't say if perfect as some  
NaN's are disturbing my program.  
```
powd64( 9.99509773402153700E+14,  -8.07157546418230000E-01 ) : -NAN 
Wolfram|alpha result :                                        7.81283570857672747453810796... × 10^-13 
Long double result rounded to _Decimal64 :                    7.812835708576727E-13 
powd64( 9.99809836120022000E+11,  -6.23247928248448100E-01 ) : -NAN 
Wolfram|alpha result :                                        3.31952801470117173393099147... × 10^-8 
Long double result rounded to _Decimal64 :                    3.319528014701172E-08 
```

pmur:
Looks like I was short a table entry in logd128. That is fixed in the PR now too.

newbie-02:
looks better now, in the limited scope of my tests ... :-)

PR #203, opened 2024-02-12
Rewrite nexttoward, and a couple minor cleanups

pmur:
Fix nexttoward in the same manner as nextafter.

Likewise, cleanup ignored test case results, and unused headers in nextafter.


newbie-02:
looks better now ...  :-)

PR #207, opened 2024-02-28
rewrite fmod and cleanup 128 bit integer conversions on ppc64

pmur:
The libdecnumber implementation fails in unexpected ways, and writing a native implementation isn't too hard.

Likewise, I had a change to improve signed 128 bit integer conversion on power10.


pmur:
That the implementation is correct for any pair of finite values is the most important requirement. Performance means nothing if the answer is wrong, especially for these non-transcendental functions.

This is a deceivingly non-trivial function to implement correctly.

newbie-02:
I agree with your desire for correct results! 
And not 'but', but 'in addition' it is of course important that calculations in the computer can be completed in a reasonable time. 
My naive idea is that the 'math' for calculating fmod need not differ too much between bin64 and Dec64, and the basic mathematical operations like plus, minus, times ... are significantly slower in Dec64 than in bin64. But only by factors between 1:10 and 1:100, see program below.  Accordingly, 'higher' functions that use them for calculation should also be slower by about the same order of magnitude. 
I have seen this in various situations. Not exactly, but in order of magnitude.  
With the new algorithms for powd64 and fmodd64 I observe considerably!!! larger differences, ~1000 for powd64 and between 1:50000 and 1:100000 for fmodd64, see also program below. And think whether improvements are possible. 
My first suspicion is that today's hardware is relatively fast with 64-bit wide data types, I think also with 128-bit **integer** values, but **not!** with 128-bit _Decimal128 types that you use.  
Thus - besides it overstretches my skills to code it myself - I expect that significant improvements are possible.  
 
 Program attached inline, github sucks in not allowing *.c attachments. 
   
 ```  
 // dec64_fmod_speed.c, program to demonstrate performance penalty of new version of fmodd64, 
// compile 'gcc -O2 -I /usr/local/include/dfp -o dec64_fmod_speed dec64_fmod_speed.c -ldfp -lm', 
// run with './dec64_fmod_speed', 
#define __STDC_WANT_DEC_FP__ 								// activate to use dec-FP functionality, 
#include <fenv.h> 									// and avoid complains about 'implicit declaration' of 'fe_dec_setround', 
#include <math.h> 									// reg. e.g. pow, powd64, 
#include <stdio.h> 									// reg. e.g. printf, 
 
#include <sys/time.h> 									// reg. e.g. gettimeofday() for performance timing, 
 
struct timeval start, end; 								// reg. performance timing, 

// timing functions: 
#define TIMEITd(expr, N) \
  gettimeofday(&start, NULL); \
  for(int i = 1; N >= i; i++) \
  { \
    expr; \
  } \
  gettimeofday(&end, NULL); \
	printf("%f; %.18E; %s \n", (end.tv_sec - start.tv_sec) + 0.000001 * (end.tv_usec - start.tv_usec), expr, #expr )
 
#define TIMEITDD(expr, N) \
  gettimeofday(&start, NULL); \
  for(int i = 1; N >= i; i++) \
  { \
    expr; \
  } \
  gettimeofday(&end, NULL); \
	printf("%f; %d; %.16DE; %s \n", (end.tv_sec - start.tv_sec) + 0.000001 * (end.tv_usec - start.tv_usec), N, expr, #expr )
 
int main() 
{ 
// we need some variables, defining them as 'volatile' helps - doesn't fully solve - to avoid the cheating behaviour 
// of the compiler to evaluate expressions at compile time, which is unwanted here as it produces different results, 
// and fakes a performance not appropriate to plan for real applications. 
	volatile double x1d, x2d, x3d; 
	volatile _Decimal64 x1DD, x2DD, x3DD; 
 
	printf( "testing fmodd64: \n" ); 
	x1DD = 1.234567890123456E+283DD; 
	x2DD = 1.5111111E+268DD; 
	printf( "TIMEITDD( x3DD = fmodd64( x1DD, x2DD ), 10000 ); fmodd64( 1.234567890123456E+283DD, 1.5111111E+268DD )  : \n" ); 
	TIMEITDD( x3DD = fmodd64( x1DD, x2DD ), 10000 ); 
	printf( " \n" ); 
 
	x1d = 1.234567890123456E+283; 
	x2d = 1.5111111E+268; 
	printf( "TIMEITd( x3d = fmod( x1d, x2d ), 10000 ); fmod( 1.234567890123456E+283, 1.5111111E+268 ) : \n" ); 
	TIMEITDD( x3d = fmod( x1d, x2d ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITDD( x3DD = ( x1DD + x2DD ), 10000 ); ( 1.234567890123456E+283DD + 1.5111111E+268DD )  : \n" ); 
	TIMEITDD( x3DD = ( x1DD + x2DD ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITd( x3d = ( x1d + x2d ), 10000 ); ( 1.234567890123456E+283 + 1.5111111E+268 ) : \n" ); 
	TIMEITDD( x3d = ( x1d + x2d ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITDD( x3DD = ( x1DD - x2DD ), 10000 ); ( 1.234567890123456E+283DD - 1.5111111E+268DD )  : \n" ); 
	TIMEITDD( x3DD = ( x1DD - x2DD ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITd( x3d = ( x1d - x2d ), 10000 ); ( 1.234567890123456E+283 - 1.5111111E+268 ) : \n" ); 
	TIMEITDD( x3d = ( x1d - x2d ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITDD( x3DD = ( x1DD * x2DD ), 10000 ); ( 1.234567890123456E+283DD * 1.5111111E+268DD )  : \n" ); 
	TIMEITDD( x3DD = ( x1DD * x2DD ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITd( x3d = ( x1d * x2d ), 10000 ); ( 1.234567890123456E+283 * 1.5111111E+268 ) : \n" ); 
	TIMEITDD( x3d = ( x1d * x2d ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITDD( x3DD = ( x1DD / x2DD ), 10000 ); ( 1.234567890123456E+283DD / 1.5111111E+268DD )  : \n" ); 
	TIMEITDD( x3DD = ( x1DD / x2DD ), 10000 ); 
	printf( " \n" ); 
 
	printf( "TIMEITd( x3d = ( x1d / x2d ), 10000 ); ( 1.234567890123456E+283 / 1.5111111E+268 ) : \n" ); 
	TIMEITDD( x3d = ( x1d / x2d ), 10000 ); 
	printf( " \n" ); 
} 
```

**Comment by pmur** [comment]
The bid implementation of __get_digits_d128 is *really* slow. That's an important function for ldexp. 

On ppc64le, it is several orders of magnitude faster (actual dfp hardware and such makes this much faster, but not almost free like bfp).

The libdecnumber version I wrote for testing might be better suited here given the deficiencies of the BID format.

**Comment by pmur** [comment]
I swapped out BID with the libdecnumber version I wrote for testing. That is probably as fast as BID is going to get with this implementation.

**Comment by newbie-02** [comment]
See factor 100 speedup in calculating 64k random pairs as well,  
see results same as before, :-)  
I don't think that's the final stage, as it's still a factor of 1000 slower than bin64 ( test program above with TIMEIT ), which is a factor of 60 with the overhead of an application program here, but it's a huge improvement.  
1. naive ideas and expectations of a 'lateral entrant' can sometimes hold, 
2. I _assume_ the new method is also based on _Decimal128, and I _assume_ it would be considerably faster if you replace it with pairs of _Decimal64, or evtl. calculations with 128-bit integers are possible?  
3. Processing speed is the! devil's foot that is being used everywhere on the net as argument against decimal data types. They cannot achieve the performance of binary data types, not on decimal optimized hardware, and certainly not on binary optimized hardware, but have the advantage of simpler I/O conversion ( think of 'SRT', grisu, ryu and so on ), and avoid most of the hassle of 'binary approximations'. IMHO they are a viable alternative, if not more than 100 times slower ( pure arithmetic ) or 5 times slower ( in the application program ) than binary.

**Comment by pmur** [comment]
My time is limited, and I think this suffices for now.  Someone could rework this to use fixed point arithmetic and get more speedup, I suspect it's always going to be 10x or so slower than bfp.  60x slower than bfp, that's pretty good all things considered.  dfp requires up to 9x more calculations per exponent difference over bfp using the same algorithm.

**Comment by newbie-02** [comment]
Understand limited time, and hope it's understood from my previous comment that I thank you very! much for what we've got so far ...

**Comment by pmur** [comment]
@stliibm can you confirm s390 is not broken by this?

**Comment by stliibm** [comment]
> @stliibm can you confirm s390 is not broken by this?

@pmur I've just successfully build master branch with
```
a9c95e6 ieee754r: use decnumber fmod implementation for BID
42f118a ieee754r: simplify fmodd128 pair-precision
c52e96e ieee754r: rewrite fmod
5fb1425 base-math: optimize for Power10 _Decimal to int128 conversions
```
No testsuite fails on s390x (64bit) / s390 (31bit) with soft/hard dfp

PR #238, opened 2025-08-07
Document this project is not maintained anymore

tuliom:
I'm afraid this commit should have been created a couple of years ago.
Life happened and the amount of time we all have to dedicate to this project reduced.
This commit would help to make it clearer the current status of the project.

@pmur , @stliibm , I'd love to read your comments on this.

----8<----

Add a note to README.developer explaining this project is not maintained anymore.
Github considers this file as the main README in its web interface.


pmur:
I have little time I can dedicate to improving libdfp. Nor have I easy and reliable access to powerpc hardware. I don't find working on soft-dfp (BID included) compelling or performant.

I've quietly held out that if anyone really used libdfp they would stumble upon the trig functions, and recoil in enough horror to open an issue. It took about 10 years, so that makes 1 user.

I am available for code reviews, and maybe keep the lights on for Fedora. I suspect most are just here for the printf/string support. The rest of the library has always felt like a first draft to claim support TR 24732.

All in all, I think the statement should be amended to welcome bug reports despite the "barely keeping the lights on" state this project is current in. They're still helpful to document all the broken glass.

newbie-02:
my two cents:  
  
there is still some demand for decimal arithmetic, esp. from people with a naive understanding of "math".  
  
It would be a good thing to satisfy their needs.  
  
With recent hardware performance shouldn't be considered a blocker anymore.  
  
The four main decimal libraries known to me: libdfp, mpdecimal, decNumber, and "IRDFPML" ( Intel and not really decimal ) all have different strengths and weaknesses. For libdfp, the strengths are integration into gcc and an easy API, while the weaknesses are inaccuracy in many cases and often alarmingly slow performance. It would/could/should be worthwhile to continue the project, as all other libraries suffer from complicated syntax ( except mpdecimal with C++? ), and the difficulties in libdfp should be largely eliminateable using the examples in the other libraries. In addition, libdfp in combination with DPD encoding can/could - I haven't tested it yet - offer extremely fast I/O, significantly faster than binary datatypes ( it works in decNumber ), thus compensating for slower calculations.  
  
That said I assume decimal types can be made competitive to binary types, even in "soft" implementation, at least to some degree, and for sure much better than today.  
  
For the points:  
A.) being able to develop for hardware supported environments too, and  
B.) IBM might be interested in contributions they could pull from further ideas developed in this project ... ???  
I'm in good hope some IBM connected people are reading here, and would expect it a small investment one might ask for, if IBM would sponsor some basic decimal powered machines to developers ( need not be "latest" nor extremely powerful, even used ones would suffice ), and or some remote access and time on bigger machines. Could pay back in multiple with even one even simple good idea a community can produce.  
  
Another consideration could be "decimal hardware for end user level is dead", however, this should not affect software implementations.
  
While we deliver as poor as today it's not to be expected to gain more attention / more use / more support, and is no surprise that python opted for mpdecimal ... improvements might change that ( attention, not python ).  
  
Perhaps we should document: "For several years now, the project has been suffering from a lack of time on the part of the maintainers. We would welcome developers who can help move it forward."

PR #273, opened 2025-11-10
Fix ceil handling of atypical 0 values, and a few other things.

pmur:
While fixing ceil, a few other issues popped up.

@stliibm I think this should address the soft-dfp bugs you'd run into. I no longer have easy access to machine to verify the non-64b targets.

### DIFF
--- a/sysdeps/s390x/convert_helpers.h
+++ b/sysdeps/s390x/convert_helpers.h
@@ -176,20 +176,10 @@ getmantd64(_Decimal64 dd, long *exp)
        : "+f" (dd) ,"=d" (tmp) );
 
   /* Convert mantissa of dd to fixed.  */
-#ifdef __s390x__
   /* On 64bit, long long int is stored in one register.  */
   asm ("cgdtr %[mant],0,%[dfp]\n\t"
        : [mant] "=d" (mantissa)
        : [dfp] "f" (dd));
-#else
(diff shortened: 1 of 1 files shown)


stliibm:
Hi @pmur,

this indeed addressed one issue (first mentioned in the already merged [pull-request](https://github.com/libdfp/libdfp/pull/272#issuecomment-3511942543)):
FAIL: test-llrint-d64
on s390x(64bit) and s390(31bit)

On s390(31bit) with sw-dfp, there still those fails:
FAIL: test-lrint-d128
FAIL: test-lrint-d64
FAIL: test-lround-d128
FAIL: test-lround-d64

On s390(31bit), the datatype long has only 4 bytes in size. According to the test-inputs lrint.input/lround.input, those tests are marked with:
`ONLY_ON_4BYTE_LONG NO_INEXACT_EXCEPTION INVALID_EXCEPTION IGNORE_RESULT`.
Currently the inexact-exception is not set, which is fine. But the invalid exception is also not set, which leads to the test errors.

The bfp variants in glibc are also raising the invalid-exception. But there is a further difference: The glibc-bfp-variants are not setting errno. The libdfp-dfp-variants are setting errno to EDOM.

pmur:
@tuliom I've included a patch to remove the (now) unsupported targets. Would you have any concerns with this patch?

tuliom (approved):

